PR #2, opened 2019-12-04
Pipeline build

christertime:
added pipeline for CI/CD to Azure Web App for Containers


christertime:
Added pipeline yaml for docker build and deploy

PR #3, opened 2020-01-09
Button component with Primary and Disabled state.

kfaln:


src/components/Select/style.ts
```diff
@@ -46,12 +47,15 @@ export type DropDownItemProps = {
 }
 
 export const DropdownItem = styled.li<DropDownItemProps>`
-    background-color: ${(props): any => props.selected ? 'var(--ui-background--light)' : 'transparent'};
+    background-color: ${(props): any => props.selected
+        ? css`${tokens.colors.ui.background__light.rgba}`
```
Cnordbo: return without interpolation.

PR #8, opened 2020-01-14
Feature/69907   button for adding new tags from pcs main

kfaln:
"Add scope" dropdown added to Preservation landing page. Currently navigating to a dummy page.


src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,14 +1,17 @@
...
 import { Select } from '../../../../components';
 import { SelectItem } from '../../../../components/Select';
+import Dropdown from '../../../../components/Dropdown';
 import Table from './../../../../components/Table';
 import { usePreservationContext } from '../../context/PreservationContext';
 
 const ScopeOverview: React.FC = (): JSX.Element => {
 
     const { project, availableProjects, setCurrentProject } = usePreservationContext();
+    const params = useParams<any>();
```
Cnordbo: Should we use path instead? That would help us avoid a dependency on a specific plant url to load the module. (moving the responsibility higher up the chain)

kfaln: Found another solution using RouteMatch... will it float?

Cnordbo: Nice

PR #9, opened 2020-01-14
Update eds-core-react to 0.0.1-alpha.9

kfaln:


Cnordbo:
Nice to read on how yarn handles versioning: 
https://yarnpkg.com/lang/en/docs/dependency-versions/

kfaln:
> As this is alpha, and doesnt comply with semver or any versioning schema for that matter, at the moment. We also have to update `package.json`. `yarn upgrade @equinor/eds-core-react --latest` might do the job?

kfaln:
`--latest` didn't do anything.. might be related to the pre-release versioning. Need to specify the exact version number to yarn in order to also update package.json.

PR #11, opened 2020-01-15
69924 component notifications toasts

arildbjerkenes:


arildbjerkenes:
Test kommer...

arildbjerkenes:
Test is added.

src/components/Notification/style.ts
```diff
@@ -0,0 +1,17 @@
...
+export const StyledSnackbarNotification = styled.div`
+    visibility: visible;
+    min-width: 250px;
+    box-shadow: 0px 3px 5px rgba(0, 0, 0, 0.2), 0px 1px 18px rgba(0, 0, 0, 0.12),
+        0px 6px 10px rgba(0, 0, 0, 0.14);
+    border-radius: 4px;
+    background-color: #333333;
+    color: #ffffff;
+    text-align: left;
+    padding: 16px;
```
Cnordbo: We use `--grid-unit` as our base for calculating spacing and margins. (EDS design grid is based on an 8px grid unit)

`padding: calc(var(--grid-unit)*2)px`

src/components/Notification/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    if (lastTimeoutNumber != -1) {
+        clearTimeout(lastTimeoutNumber);
+    }
+
+    const timeoutNumber = setTimeout(() => {
+        render(<div />, notificationContainer);
+    }, duration);
+
+    lastTimeoutNumber = timeoutNumber;
```
Cnordbo: This should be removed.

arildbjerkenes:
> The entire set of files should probably be moved to a different namespace, `core/services` ?
> 
> As this is not something that is ever referred to as a component, but rather as a library.

Fikset.

src/core/services/NotificationService/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * @param message Message to display
+ * @param duration Duration of notification, in milliseconds
+ */
+export const showSnackbarNotification = (
+    message: string,
+    duration: number
+): any => {
+    render(<Notification message={message} />, container);
+
+    clearTimeout(lastTimeoutId);
```
Cnordbo: We should clear the timeout before rendering, just to avoid our newly rendered component to be cleared by a triggered timer.

src/core/services/NotificationService/index.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { render } from 'react-dom';
+import { StyledSnackbarNotification } from './style';
+
+let lastTimeoutId = -1;
+const container = document.createElement('div');
```
Cnordbo: Give the container an ID so its easier to debug \ identify later.

src/core/services/NotificationService/style.ts
```diff
@@ -0,0 +1,17 @@
...
+    box-shadow: 0px 3px 5px rgba(0, 0, 0, 0.2), 0px 1px 18px rgba(0, 0, 0, 0.12),
+        0px 6px 10px rgba(0, 0, 0, 0.14);
+    border-radius: 4px;
+    background-color: #333333;
+    color: #ffffff;
+    text-align: left;
+    padding: calc(var(--grid-unit) * 2);
+    position: fixed;
+    z-index: 1;
+    left: 2%;
```
Cnordbo: use calc(var(--grid-unit) * 2) - or something thereof ( look at spacing in designs )

src/core/services/NotificationService/style.ts
```diff
@@ -0,0 +1,17 @@
...
+    bottom: 30px;
```
Cnordbo: use `calc(var(--grid-unit) * 2)` - or something thereof ( look at spacing in designs )

src/index.tsx
```diff
@@ -44,10 +43,7 @@ const start = async (): Promise<void> => {
     element.setAttribute('id', 'app-container');
     document.body.appendChild(element);
 
-    ReactDOM.render(
-        <Root />,
-        document.getElementById('app-container'));
-
+    ReactDOM.render(<Root />, document.getElementById('app-container'));
 };
```
Cnordbo: No actual changes, revert.

src/components/index.ts
```diff
@@ -1,3 +1,3 @@
-export {default as Dropdown} from './Dropdown/index';
-export {default as Loading} from './Loading/index';
-export {default as Select} from './Select/index';
+export { default as Dropdown } from './Dropdown/index';
+export { default as Loading } from './Loading/index';
+export { default as Select } from './Select/index';
```
Cnordbo: no changes, revert.

src/core/services/NotificationService/style.ts
```diff
@@ -0,0 +1,17 @@
+import styled from 'styled-components';
+
+export const StyledSnackbarNotification = styled.div`
+    visibility: visible;
```
Cnordbo: Is this a default value? Can it be removed?

src/core/services/NotificationService/style.ts
```diff
@@ -0,0 +1,17 @@
...
+    min-width: 250px;
+    box-shadow: 0px 3px 5px rgba(0, 0, 0, 0.2), 0px 1px 18px rgba(0, 0, 0, 0.12),
+        0px 6px 10px rgba(0, 0, 0, 0.14);
+    border-radius: 4px;
+    background-color: #333333;
+    color: #ffffff;
+    text-align: left;
```
Cnordbo: Is this a default value? Can it be removed?

src/core/services/NotificationService/style.ts
```diff
@@ -0,0 +1,17 @@
...
+    padding: calc(var(--grid-unit) * 2);
+    position: fixed;
+    z-index: 1;
```
Cnordbo: We should make sure this needs to be present, or remove it.

src/core/services/NotificationService/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const showSnackbarNotification = (
+    message: string,
+    duration: number
+): any => {
+    render(<Notification message={message} />, container);
+
+    clearTimeout(lastTimeoutId);
+
+    lastTimeoutId = setTimeout(() => {
+        render(<div />, container);
```
Cnordbo: try rendering `false`, instead of an empty dom element.

src/core/services/NotificationService/style.ts
arildbjerkenes: Not sure what you mean. Later we will have styles for dialog- and top notification.

src/core/services/NotificationService/style.ts
arildbjerkenes: We will update this later when get exact design from EDS.

src/core/services/NotificationService/style.ts
arildbjerkenes: I suggest we leave this for now. We will receive updates to this later.

src/core/services/NotificationService/index.tsx
arildbjerkenes: False was not possible, but we can use react fragments to do this.

src/core/services/NotificationService/style.ts
Cnordbo: Im referring to the `visibility: visible;` property, not the component.

src/core/services/NotificationService/style.ts
arildbjerkenes: z-index decides the stack order of elements. We would like notification to be on top. If no one else use higher z-index, then 1 will be ok.

src/core/services/NotificationService/style.ts
arildbjerkenes: text-left can be removed.

src/core/services/NotificationService/style.ts
arildbjerkenes: You right, we don't visibility anymore.

src/core/services/NotificationService/style.ts
Cnordbo: Its not if `1` is OK or not, the question is if its needed. If you dont know why the property is there, and the feature functions as intended without it, then it should be removed.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -7,72 +7,89 @@ import { SelectItem } from '../../../../components/Select';
...
 
     useMemo(async () => {
         const tags = await apiClient.getTags();
         console.log('Tags: ', tags);
-    },[]);
+    }, []);
 
     const changeProject = (project: SelectItem): void => {
         setCurrentProject(project.value as number);
+        showSnackbarNotification('æest' + project.value, 3000);
```
Cnordbo: This looks like it should be removed?

PR #13, opened 2020-01-21
Feature/69912 add scope   set tag properties

Cnordbo:
* Refactors the Select and Dropdown components
* Adds a "Step 2" to Add Scope, with mocked data.


src/modules/Preservation/views/AddScope/SetTagProperties/SetTagProperties.tsx
```diff
@@ -0,0 +1,102 @@
...
+                        data={stepsDataMapped}
+                        label={'Preservation step'}
+                    >
+                        {(step > -1 && steps[step].text) || 'Select step'}
+                    </SelectInput>
+                </InputContainer>
+                <InputContainer>
+                    <TextField
+                        id={'Remark'}
+                        style={{ maxWidth: '480px' }}
```
kfaln: Should we override styling of the TextField component? Consider wrapping it in another element with custom style.

Cnordbo: I see your point, i am just not sure if we want to "componetize" ourself in such detail. Just wrapping sub-components to enforce a width of individual elements. This adds complexity to the dom by introducing a bunch of extra nodes, just there for the sake of styling ( which is fine, but should be reduced as much as possible ).  
If all inputs had the same width, i would definetly had a wrapper for the entire form instead setting the explicit width on each element.

kfaln:
@Cnordbo - I have merged the stuff. You may complete this one :)

PR #14, opened 2020-01-22
Add scope - handle data in parent + initial tag search

kfaln:


src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -118,8 +124,29 @@ class PreservationApiClient extends ApiClient {
             id: 2
         }]);
     }
-}
 
+    async getTagsForAddPreservationScope(tagNo: string, setRequestCanceller?: RequestCanceler): Promise<TagData[]> {
+        // const endpoint = '/Tags';
```
Cnordbo: `/Tags/Search`

We also need to setup support for sending in Project that should be set as query string.

src/modules/Preservation/views/AddScope/types.ts
```diff
@@ -1,3 +1,9 @@
 export type Tag = {
     id: number;
 }
+
+export type TagRow = {
+    tagId: number;
+    tagNo: string;
+    description: string;
+}
```
Cnordbo: Only needed by Table in SelectTags, can be removed from common types.

src/modules/Preservation/views/AddScope/SelectTags.tsx
```diff
@@ -1,69 +1,46 @@
...
 type SelectTagsProps = {
-    tags: Tag[];
+    selectedTags: Tag[];
+    scopeTableData: TagRow[];
     setSelectedTags: (tags: Array<Tag>) => void;
+    searchTags: (tagNo: string | null) => void;
     nextStep: () => void;
 }
 
-type TagRow = {
```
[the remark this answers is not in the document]
kfaln: It's actually used in both AddScope (where the data resides in state) and SelectTags (where the data is presented in Table). I'm gonna hold on to it a while longer in common types.

src/modules/Preservation/http/PreservationApiClient.ts
kfaln: Yes, will look into this next.

PR #16, opened 2020-01-24
Tag list and start preservation

arildbjerkenes:
Because of dependencies, I have implemented the tag list as well, in this branch. 

There are some fields missing from the endpoint. Will add this later. 

Note: After starting preservation on a tag, we call the endpoint to get the tags. It seems like the backend returns a list that is not updated (caching issue?). Will look into this.


arildbjerkenes:
There are some merge conflicts. I will resolve these now.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,78 +1,155 @@
...
-    useMemo(async () => {
-        const tags = await apiClient.getTags();
-        console.log('Tags: ', tags);
-    },[]);
+    async function getPreservedTags(): Promise<void> {
+        const tags = await apiClient.getPreservedTags();
+        setTags(tags);
+    }
+
+    useMemo(getPreservedTags, []);
```
Cnordbo: `useEffect` or assign result of memo, instead of setting tags in state. 

`tags = useMemo(getPreservedTags, [])`

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,78 +1,155 @@
...
+import PlayArrowOutlinedIcon from '@material-ui/icons/PlayArrowOutlined';
+import CompareArrowsOutlinedIcon from '@material-ui/icons/CompareArrowsOutlined';
+import CreateOutlinedIcon from '@material-ui/icons/CreateOutlined';
+import DeleteOutlinedIcon from '@material-ui/icons/DeleteOutlined';
+import PrintOutlinedIcon from '@material-ui/icons/PrintOutlined';
+import IconButton from '@material-ui/core/IconButton';
+import { PreservedTag } from './types';
 
 const ScopeOverview: React.FC = (): JSX.Element => {
+    const [startPreservationDisabled, setStartPreservationDisabled] = useState(
```
Cnordbo: `useMemo` instead of useState. 
UseMemo can automatically recalculate itself everytime `selectedTags` is changed.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,78 +1,155 @@
...
-                selection: true
-            }}
-            onSelectionChange={(test): void => console.log('Selection changed: ', test)}
-            style={{ boxShadow: 'none' }} />
+            });
+        }
+        setStartPreservationDisabled(!enableStartPreservation);
+        console.log('Set selected tags: ' + selected);
+        setSelectedTags(selected);
+    };
```
Cnordbo: See comment above regarding using `useMemo` instead of state for handling enable\disable Start preservation. 

Suggestion: remove method

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,78 +1,155 @@
...
+                    { title: 'Resp', field: 'responsible' },
+                    { title: 'Disc', field: 'disciplineCode' },
+                    { title: 'Status', field: 'status' },
+                ]}
+                data={tags}
+                options={{
+                    showTitle: false,
+                    selection: true,
+                }}
+                onSelectionChange={onSelectionHandler}
```
Cnordbo: Can we use `setSelectedTags` directly here?

src/modules/Preservation/views/ScopeOverview/types.ts
```diff
@@ -0,0 +1,25 @@
...
+    stepId: string;
+    tagFunctionCode: string;
+    needUserInput: string;
+    //requirements: Requirements[]; //todo
+}
+
+export interface Requirements {
+    nextDueTimeUtc: string;
+    nextDueAsYearAndWeek: string;
+}
```
Cnordbo: I don\`t think we should extract types \ interfaces into `types.ts` before the same type is used somewhere else in the same module.

arildbjerkenes:
Conflicts is solved.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
arildbjerkenes: So the onSelectionHandler will only create an array with selected tags, and call setSelectedTags, and then a useMemo will depend on selectedTags, and will traverse the array, to find out if the button can be enabled? 

Then we will traverse the array twice?  Why is this better?

src/modules/Preservation/views/ScopeOverview/types.ts
arildbjerkenes: You mean it should be in ScopeOverview, as long as it is not used in other scripts?

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
Cnordbo: Separation of logic, a lot easier to manage and read. Reduces the number of side effects from a single function, and its likely a non-issue performance wise. 

```js
onSelectionHandler = (selectedTags): void => {
    setSelectedTags(() => selectedTags.map(t => t.id));
}

startPreservationEnabled = useMemo(() => selectedTags.length > 0 && selectedTags.every(t => t.status === 'NotStarted'), [selectedTags]) 
```

or something similar.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -15,78 +33,118 @@ const ScopeOverview: React.FC = (): JSX.Element => {
...
     };
 
+    const startPreservation = (): void => {
+        console.log('Start preservation for selected tags.' + selectedTags);
+        apiClient.startPreservation(selectedTags);
+        getPreservedTags();
+        setStartPreservationDisabled(true); //after rendring, tags are unselected, but onSelectionChange is not called
+    };
+
+    const onSelectionHandler = (selectedTags: any): void => {
```
Cnordbo: SelectedTags should probably not be any? seems to be a `PreservedTag[]`

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,12 +1,30 @@
...
+} from './ScopeOverview.style';
 import Dropdown from '../../../../components/Dropdown';
 import Table from './../../../../components/Table';
 import { usePreservationContext } from '../../context/PreservationContext';
+import PlayArrowOutlinedIcon from '@material-ui/icons/PlayArrowOutlined';
+import CompareArrowsOutlinedIcon from '@material-ui/icons/CompareArrowsOutlined';
+import CreateOutlinedIcon from '@material-ui/icons/CreateOutlined';
+import DeleteOutlinedIcon from '@material-ui/icons/DeleteOutlined';
+import PrintOutlinedIcon from '@material-ui/icons/PrintOutlined';
+import IconButton from '@material-ui/core/IconButton';
```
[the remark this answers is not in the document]
Cnordbo: @kfaln To use this option - we just need to verify our bundling strategy - i think i fixed this earlier, but if you got time to verify, please do. 

https://material-ui.com/components/icons/#usage

If we use this approach and don`t verify our bundling strategy, then all icons of the entire material-ui will be included in the bundle.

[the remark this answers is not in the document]
kfaln: Looking at the config I don't think this is configured (something with Babel). So never mind my comment at this time, it's probably best to leave it as is.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
arildbjerkenes: no

Cnordbo:
Because of auto-formatting, your need to re-merge with master and fix any code. 
An example would be in `PreservationApiClient` - Old code appears assnewly added code in your commit and would add boilerplate code back into the codebase.

src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -77,67 +117,105 @@ class PreservationApiClient extends ApiClient {
...
+            {
+                text: 'Journey 1',
+                id: 1,
+            },
+            {
+                text: 'Journey 2',
+                id: 2,
+            },
+        ]);
     }
```
Cnordbo: This is removed in master, but re-added here due to auto-formatting.

src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -77,67 +117,105 @@ class PreservationApiClient extends ApiClient {
...
+            {
+                text: 'Step 1',
+                id: 1,
+            },
+            {
+                text: 'Step 2',
+                id: 2,
+            },
+        ]);
     }
```
Cnordbo: This is removed in master, but re-added here due to auto-formatting.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,92 +1,203 @@
...
     }, []);
 
     const changeProject = (event: React.MouseEvent, index: number): void => {
         event.preventDefault();
         setCurrentProject(availableProjects[index].id);
     };
 
+
+    const startPreservation = (): void => {
+        console.log('Start preservation for selected tags.', selectedTags);
```
Cnordbo: Remove console.log

src/modules/Preservation/http/PreservationApiClient.ts
arildbjerkenes: master look to be the same as I have.

src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -77,29 +117,52 @@ class PreservationApiClient extends ApiClient {
...
-        const endpoint = '/Tags';
-        const settings: AxiosRequestConfig = {};
+    async getPreservedTags(projectName: string,
+        setRequestCanceller?: RequestCanceler
+    ): Promise<PreservedTagResponse[]> {
+        const endpoint = '/Tags/Preserved';
+
+        const settings: AxiosRequestConfig = {
+            params: {
+                projectName: 'Project-16',
```
kfaln: Hardcoded project name - intentional?

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,92 +1,215 @@
...
+        nextDueWeeks: string;
+    };
+    requirements: {
+        id: string;
+        requirementDefintionId: string;
+        nextDueTimeUtc: string;
+        nextDueAsYearAndWeek: string;
+        nextDueWeeks: string;
+    }[];
+    tableData: {
```
kfaln: May not need the `tableData` property here and in the `PreservedTagResponse` interface in PerservationApiClient? Ended up not needing to manipulate the value?

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,92 +1,215 @@
...
                 options={{
                     showTitle: false,
                     selection: true,
+                    pageSize: 10,
+                    pageSizeOptions: [10, 50, 100],
+                    headerStyle: {
+                        backgroundColor: '#f7f7f7'
+                    },
+
+                    rowStyle: rowData => {
```
kfaln: Tip on how to shorthand this a bit (I think this will work :))

                    rowStyle: (rowData): any => ({
                        color: rowData.firstUpcomingRequirement?.nextDueWeeks < 0 && '#B30D2F'
                    })

src/modules/Preservation/http/PreservationApiClient.ts
Cnordbo: `git fetch` & `git merge origin master`

src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -5,19 +5,54 @@ import { RequestCanceler } from '../../../http/HttpClient';
...
+        id: string;
+        requirementDefintionId: string;
+        nextDueTimeUtc: string;
+        nextDueAsYearAndWeek: string;
+        nextDueWeeks: string;
+    }[];
+    tableData: {
+        checked: boolean;
+    };
+}
```
Cnordbo: This has a bunch of properties that does not align with the API response. 

UI specific data models should only be created where used. 

```js
  {
    "areaCode": "string",
    "calloffNo": "string",
    "commPkgNo": "string",
    "description": "string",
    "disciplineCode": "string",
    "id": 0,
    "isAreaTag": true,
    "isVoided": true,
    "mcPkgNo": "string",
    "purchaseOrderNo": "string",
    "requirements": [
      {
        "nextDueTimeUtc": "2020-01-29T08:26:15.310Z",
        "nextDueAsYearAndWeek": "string"
      }
    ],
    "status": "NotStarted",
    "stepId": 0,
    "tagFunctionCode": "string",
    "tagNo": "string",
    "needUserInput": true
  }
```

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
arildbjerkenes: worked :-)

src/modules/Preservation/views/ScopeOverview/__tests__/ScopeOverview.test.jsx
```diff
@@ -0,0 +1,30 @@
...
+    });
+
+    it('Start preservation button should be disabled when Active tags are selected', () => {
+        //todo
+    });
+
+    it('Start preservation button should be enabled when NotStarted tags are selected', () => {
+        //todo
+    });
+});
```
Cnordbo: Invalid test - should be completed before review is requested.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,92 +1,209 @@
...
         setCurrentProject,
         apiClient,
     } = usePreservationContext();
 
-    useMemo(async () => {
-        const tags = await apiClient.getTags();
-        console.log('Tags: ', tags);
+    const getTags = async (): Promise<void> => {
+        setIsLoading(true);
+        const tags = await apiClient.getPreservedTags(project.name);
```
Cnordbo: This probably needs some additional mapping after receiving the response. (Converting to a valid Date object)

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -1,92 +1,209 @@
...
+     * if there are selected rows with other status than NotStarted
+     */
+    useEffect(
+        () => {
+            setStartPreservationDisabled(
+                selectedTags.length === 0 ||
+                selectedTags.findIndex((t) => t.status !== 'NotStarted') !== -1
+            );
+        }
+    );
```
Cnordbo: Missing requirements parameter `[selectedTags]`

PR #17, opened 2020-01-27
Add scope: Tuning to table rendering and options

kfaln:
- "Preserved" column with checkbox icon
- Table header style
- Toolbar component (number of rows selected)
- Disable selection of preserved tags (see bug comment)

### DIFF
--- a/src/modules/Preservation/views/AddScope/SelectTags.tsx
+++ b/src/modules/Preservation/views/AddScope/SelectTags.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { Button, TextField } from '@equinor/eds-core-react';
-import CheckBoxIcon from '@material-ui/icons/Checkbox';
+import CheckBoxIcon from '@material-ui/icons/CheckBox';
 import { Tag, TagRow } from './types';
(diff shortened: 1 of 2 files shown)


src/modules/Preservation/views/AddScope/SelectTags.tsx
```diff
@@ -67,8 +72,20 @@ const SelectTags = (props: SelectTagsProps): JSX.Element => {
...
+                        headerStyle: {
+                            backgroundColor: '#f7f7f7'
+                        },
                         selection: true,
-                        search: false
+                        selectionProps: (data: TagRow): any => ({
+                            disabled: data.isPreserved,
+                            // Bug: 'Select all' will also select disabled checkboxes: https://github.com/mbrn/material-table/issues/686
+                            // Disabled checkbox should be hidden, but that would also hide the 'select all' problem
+                            // style: { display: rowData.isPreserved && 'none' }
```
Cnordbo: Create PBI to fix this issue - potential workaround could be to hook into the onChange event of the checkbox and remove the check again if checked === true && disabled === true.

kfaln: Have created the PBI. I have already tested the workaround. It kind of works, but it breaks the "deselect all" functionality afterwards.

PR #20, opened 2020-01-29
Add scope - Set properties - Selected tags "sidemodule"

kfaln:
- List of selected tags when setting properties for "add scope".
- Added header to "SetTagProperties".


src/modules/Preservation/views/AddScope/AddScope.style.ts
```diff
@@ -0,0 +1,24 @@
...
+    display: flex;
+`;
+
+export const TagProperties = styled.div`
+    width: 70%;
+`;
+
+export const SelectedTags = styled.div`
+    width: 30%;
+    margin-left: calc(var(--grid-unit));
```
Cnordbo: no need for calc when not doing any calculation

PR #31, opened 2020-02-11
Tag details "flyout"/sidebar with header

kfaln:


src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -0,0 +1,130 @@
...
+    setDisplayFlyout: (displayFlyout: boolean) => void;
+    tagNo: string;
+    tagId: number;
+}
+
+const TagFlyout = ({
+    displayFlyout,
+    setDisplayFlyout,
+    tagNo,
+    tagId
```
Cnordbo: We only need the tagID to show this view, which will give us the tagNo when data is loaded. I can see the value in displaying the tagno while the data is loading - but it can lead to mismatching between tagno and tagdetails. (Current implementation allows changing tagNo, but it still fetches data for old TagID if not changed)

Suggestion is to just leave a data loading spinner while fetching all data for tagID.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+interface TagFlyoutProps {
+    displayFlyout: boolean;
+    setDisplayFlyout: (displayFlyout: boolean) => void;
+    tagNo: string;
+    tagId: number;
+}
+
+const TagFlyout = ({
+    displayFlyout,
```
Cnordbo: Should the flyout just display as soon as its rendered instead? This avoids a bunch of re-renders on the flyout, as well as effects having to check if they should update something based on `displayFlyout` property.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -0,0 +1,130 @@
...
+        }
+    };
+
+    if (displayFlyout) {
+        return (
+            <Container onMouseDown={(): void => setDisplayFlyout(false)}>
+                <Flyout ref={flyoutRef} onMouseDown={(event: MouseEvent): void => event.stopPropagation()}>
+                    <FlyoutHeader>
+                        <h1>{tagNo}</h1>
+                        <StatusLabel status={tagDetails?.status}>
```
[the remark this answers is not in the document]
kfaln: I was not aware of the limited browser support, so that's nice to know. I have moved away from the usage of optional chaining where I could find it (in code related to this pull request, that is).

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const { apiClient } = usePreservationContext();
+
+    // fade-in effect
+    useEffect((): void => {
+        setTimeout((): void => {
+            if (flyoutRef.current) {
+                flyoutRef.current.style.opacity = '1';
+            }
+        }, 1);
+    }, [displayFlyout]);
```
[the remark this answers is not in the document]
kfaln: `displayFlyout` is gone (see other comment) and replaced with `flyoutRef`. The timeout seems to be required, however, in order for the transition to kick in (it needs to go "from something to something else" just slightly after the component is rendered (0 to 1 in this case). Otherwise it's happening too fast and no "animation" effect will be visible.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: Yeah, it was just to have some info to display while loading the rest. Was considering removing `tagNo`. You convinced me :)

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -52,6 +54,9 @@ const ScopeOverview: React.FC = (): JSX.Element => {
     const [tags, setTags] = useState<PreservedTag[]>([]);
     const [selectedTags, setSelectedTags] = useState<PreservedTag[]>([]);
     const [isLoading, setIsLoading] = useState<boolean>(false);
+    const [displayFlyout, setDisplayFlyout] = useState<boolean>(false);
+    const [flyoutTagNo, setFlyoutTagNo] = useState<string>('');
+    const [flyoutTagId, setFlyoutTagId] = useState<number>(0);
```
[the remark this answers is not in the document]
kfaln: Yup.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: I have changed this to conditional rendering now and removed the passing and checking of `displayFlyout` (see `ScopeOverview.tsx`). I like it better, is that what you meant?

kfaln:
Will try out the extract to separate component stuff...

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: The Flyout container is now also moved to a separate component.

kfaln:
Moved Flyout container to a separate component (for potential re-usability later on).

PR #32, opened 2020-02-12
Feature/71252 ui create new area tag and add to scope

arildbjerkenes:


src/modules/Preservation/views/AddScope/AddScope.tsx
```diff
@@ -135,14 +178,34 @@ const AddScope = (): JSX.Element => {
...
         case 1:
-            return <SelectTags
-                nextStep={goToNextStep}
-                setSelectedTags={setSelectedTags}
-                searchTags={searchTags}
-                selectedTags={selectedTags}
-                scopeTableData={scopeTableData}
-                isLoading={isLoading}
-            />;
+            if (props.match.params.method === 'selectTags') {
```
kfaln: We could strongly type this a bit more perhaps by using an Enum and a helper function. Something like this:

    enum AddScopeMethod {
        AddTagsManually,
        CreateAreaTag,
        Unknown
    }

    const getAddScopeMethod = (): AddScopeMethod => {
        switch (props.match.params.method) {
            case 'selectTags':
                return AddScopeMethod.AddTagsManually;
            case 'createAreaTag':
                return AddScopeMethod.CreateAreaTag;
            default:
                return AddScopeMethod.Unknown;
        }
    };

And then use it in the `return` function:

    const addScopeMethod = getAddScopeMethod();

    switch (step) {
        case 1:
            if (addScopeMethod === AddScopeMethod.AddTagsManually) {
                return <SelectTags
                    ...
                />;
            }
            else if (addScopeMethod === AddScopeMethod.CreateAreaTag) {
                return <CreateAreaTag
                    ...
                />;
            }
            else {
                return <div>Unknown method</div>;
            }

src/modules/Preservation/views/AddScope/AddScope.tsx
```diff
@@ -154,7 +217,7 @@ const AddScope = (): JSX.Element => {
                             journeys={journeys}
                             requirementTypes={requirementTypes}
                             previousStep={goToPreviousStep}
-                            submitForm={submitForm}
+                            submitForm={submitFormTags}
```
kfaln: I feel it would be better with two separate submit functions. A function should ideally only do "one thing".

Expanding on the previous comment, you could distinguish them with the enum:
`submitForm={addScopeMethod === AddScopeMethod.AddTagsManually ? submitAddTagsManually : submitCrateAreaTag}`

src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const [mappedDisciplines, setMappedDisciplines] = useState<SelectItem[]>([]);
+    const [mappedAreas, setMappedAreas] = useState<SelectItem[]>([]);
+
+    const freetextInputRef = useRef<HTMLInputElement>(null);
+    const descriptionInputRef = useRef<HTMLInputElement>(null);
+
+    //Build tagNo
+    let newTagNo = '';
+    if (props.areaType && props.discipline && props.description) {
+        newTagNo = props.areaType.value + '-' + props.discipline.code;
```
kfaln: Could use string interpolation instead of concatenation (here and other places):

```
newTagNo = `${props.areaType.value}-${props.discipline.code}`;
```

src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
```diff
@@ -0,0 +1,168 @@
...
+                        >
+                            {(props.discipline != undefined && props.discipline.description) || 'Select'}
+                        </SelectInput>
+                    </FormFieldSpacer>
+                    <FormFieldSpacer>
+                        <SelectInput
+                            onChange={setAreaForm}
+                            data={mappedAreas}
+                            label={'Area'}
+                            hidden={props.areaType === undefined || props.areaType.value !== '#PRE'}
```
kfaln: The `Select` component should not be responsible for determining whether to render it self or not by using a `hidden` prop. The component "hosting" the Select should be in charge of this.

Recommend using a "conditional" rendering, something like this (you will need to verify the reversed logic):

                    <FormFieldSpacer>
                        {
                            (props.areaType !== undefined && props.areaType.value === '#PRE') && 
                            <SelectInput
                                onChange={setAreaForm}
                                data={mappedAreas}
                                label={'Area'}
                            >
                                {(props.area != undefined && props.area.description) || 'Type to select'}
                            </SelectInput>
                        }
                    </FormFieldSpacer>

The `hidden` prop can then be removed.

src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
```diff
@@ -0,0 +1,168 @@
...
+    return (
+        <div>
+            <Header>
+                <h1>Create Area Tag</h1>
+                <div>{project.description}</div>
+            </Header>
+            <Container>
+                <InputContainer>
+                    <FormFieldSpacer>
+                        <SelectInput
```
kfaln: This component is missing some styling that you could use this opportunity to add:
- backgound color on item hover
- "pointer" cursor (hand) on item hover

See the dropdowns in header menu for reference.

src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -325,6 +390,43 @@ class PreservationApiClient extends ApiClient {
...
     }
+
+    /**
+     * Get disciplines
+     *
+     * @param setRequestCanceller Returns a function that can be called to cancel the request
+     */
+    async getDisciplines(setRequestCanceller?: RequestCanceler): Promise<DisciplineResponse[]> {
+
+        //const endpoint = '/Disciplines';
```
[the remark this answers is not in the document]
arildbjerkenes: Endpoint is now in place, so I have updated the client.

src/modules/Preservation/views/AddScope/AddScope.tsx
```diff
@@ -75,19 +111,26 @@ const AddScope = (): JSX.Element => {
...
+    const submitFormTags = async (stepId: number, requirements: Requirement[], remark: string | null): Promise<void> => {
         try {
-            const listOfTagNo = selectedTags.map(t => t.tagNo);
-            await apiClient.preserveTags(listOfTagNo, stepId, requirements, project.name, remark);
-            showSnackbarNotification(`${listOfTagNo.length} tags successfully added to scope`, 5000);
-            history.push('/');
+            if (areaType) {
+                //Preserver a new area tag
+                const tagNo = selectedTags[0].tagNo;
+                await apiClient.preserveNewAreaTag(tagNo, areaType?.value, discipline?.code, area?.code, freetext, stepId, requirements, project.name, remark);
```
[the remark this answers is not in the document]
arildbjerkenes: We are using this other places as well. Not only by my code ;-).  Must be fixed all over.

src/modules/Preservation/views/AddScope/AddScope.tsx
```diff
@@ -23,6 +25,14 @@ const AddScope = (): JSX.Element => {
     const [requirementTypes, setRequirementTypes] = useState<RequirementType[]>([]);
     const [isLoading, setIsLoading] = useState(false);
 
+    const [disciplines, setDisciplines] = useState<Discipline[]>([]);
+    const [areas, setAreas] = useState<Area[]>([]);
+    const [areaType, setAreaType] = useState<SelectItem | undefined>();
```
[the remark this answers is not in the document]
arildbjerkenes: I can have an array with strings, and then a mapping to SelectItems, but why should I have that?

src/modules/Preservation/views/AddScope/AddScope.tsx
```diff
@@ -23,6 +25,14 @@ const AddScope = (): JSX.Element => {
...
+    const [discipline, setDiscipline] = useState<Discipline | undefined>();
```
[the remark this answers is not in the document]
arildbjerkenes: Yepp

src/modules/Preservation/views/AddScope/AddScope.tsx
arildbjerkenes: done!

src/modules/Preservation/views/AddScope/AddScope.tsx
arildbjerkenes: I realize now that this would be rendered when we know the method.

src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
arildbjerkenes: Old habbit.

src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
arildbjerkenes: Agree.

src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
arildbjerkenes: I prefer that we create a PBI, and include all findings on styling issues. We might get some input from Lotte as well.  

71430: UI - Fix findings on styling

src/modules/Preservation/views/AddScope/AddScope.tsx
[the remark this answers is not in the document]
kfaln: In the context for creating an Area Tag (submit), you only need the value from `areaType`, which is a string. It seems strange to use a type from the Select component, which also includes other fields, when you only need one of them (`value`). 

The point is, you could probably just set the string value in a state variable when doing the action in `CrateAreaTag` and then use this directly when submitting the form. Then we could also remove the import of `SelectItem` in `AddScope`.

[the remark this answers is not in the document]
kfaln: This also applies for `discipline` and `area`. We just need the code, not the type...

src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -231,6 +242,60 @@ class PreservationApiClient extends ApiClient {
...
+        this.setupRequestCanceler(settings, setRequestCanceller);
+        try {
+            await this.client.post(endpoint, {
+                tagNo: tagNo,
+                areaType: areaType,
+                disciplineCode: disciplineCode,
+                areaCode: areaCode,
+                tagNoFreetext: freetext,
+                projectName: projectName,
+                stepId: stepId,
```
Cnordbo: shorthand can be used for all properties, but should we be explicit about it? I prefer being as clear as possible, especially when interacting with external resources. So maybe the above mentioned lines are the way we should do it - but the point is in any case so be consistent. (Dont mix shorthand non-shorthand properties).

src/modules/Preservation/views/AddScope/AddScope.tsx
```diff
@@ -1,28 +1,44 @@
...
+import CreateAreaTag from './CreateAreaTag/CreateAreaTag';
 import Spinner from '../../../../components/Spinner';
 import TagDetails from './TagDetails/TagDetails';
 import { showSnackbarNotification } from './../../../../core/services/NotificationService';
 import { useHistory } from 'react-router-dom';
 import { usePreservationContext } from '../../context/PreservationContext';
+import { SelectItem } from '../../../../components/Select';
 
-const AddScope = (): JSX.Element => {
+const AddScope = (props: any): JSX.Element => {
```
Cnordbo: Dont use `any`

src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
```diff
@@ -0,0 +1,168 @@
...
+    areaType: SelectItem | undefined;
+    setAreaType: (areaType: SelectItem | undefined) => void;
+    discipline: Discipline | undefined;
+    setDiscipline: (discipline: Discipline | undefined) => void;
+    area: Area | undefined;
+    setArea: (area: Area | undefined) => void;
+    freetext: string | undefined;
+    setFreetext: (freetext: string | undefined) => void;
+    description: string | undefined;
+    setDescription: (description: string | undefined) => void;
```
Cnordbo: use optional or require the property where possible.

src/modules/Preservation/http/PreservationApiClient.ts
arildbjerkenes: Not sure what kind of shorthand you refer to.

src/modules/Preservation/views/AddScope/AddScope.tsx
arildbjerkenes: I have created a props type, but I must use any to define 'match'. I don't think there is any type defined for this.

arildbjerkenes:
> This approach fetches a bunch of resources that will never be used based on which module your loading. (createAreaTag and manual tag scoping) - which in effect has a performance hit on the environment.
> We should probably re-think the strategy used here and separate the CreateAreaTag and AddTag modules.

arildbjerkenes:
We

arildbjerkenes:
AddScope will as Christer pointed out, fetch Disciplines and Areas even when adding standard tags. My plan was to include useEffects for these in CreateAreaTag. However, then these will be loaded every time you reentered the create area tag screen. We don't have time for refactoring of this now. Suggest we create a PBI to refactor this later.

PR #35, opened 2020-02-19
71184 ui transfer button with notification and spinner

arildbjerkenes:
The first version of transfer-button will have a simplified dialog box. When we have proper design, we will probably create a modal dialog component, that can take a child component.


src/core/services/ModalDialogService/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    };
+
+    window.onclick = (e: any): void => {
+        if (e.target.id === 'Modal') {
+            cancel();
+        }
+    };
+
+    return (
+        <Modal id='Modal'>
```
Cnordbo: Nitpicking - But this just threw me off a bit. What about changing it to `ModalOverlay` as the modal itself is basically what you put in `ModalContent`

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -82,46 +84,63 @@ const ScopeOverview: React.FC = (): JSX.Element => {
...
-                                'Selected tags have been preserved for this week.',
-                                5000
-                            );
-                        }
-                    );
-                }
-            );
+            await apiClient.transfer(selectedTags.map(t => t.id));
+            setSelectedTags([]);
+            getTags().then(() => {
```
Cnordbo: Async method, use `await` for better looking code.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -82,46 +84,63 @@ const ScopeOverview: React.FC = (): JSX.Element => {
...
+            console.error('Transfer failed: ', error.messsage, error.data);
+            showSnackbarNotification(error.message, 5000);
+        }
+        setIsLoading(false);
+        return Promise.resolve();
+    };
+
+    const transferDialog = (): void => {
+        //Verify that all selected tags can be transfered
+        let numTagsNotTransferable = 0;
```
Cnordbo:
```js
const numTagsNotTransferable = selectedTags.filter((tag) => !tag.readyToBeTransferred);
```

PR #37, opened 2020-02-21
Tag Requirements - Save

kfaln:
* Save user input on tag requirements (number + checkbox) and comments.
* Enable/disable logic on buttons.
* Readonly logic when preservation has not started.
* Preservation not started notification in header.
* Allow snackbar notification to display on right-hand side (via props, for Flyout).

### DIFF
--- a/src/components/Checkbox/index.tsx
+++ b/src/components/Checkbox/index.tsx
@@ -4,19 +4,26 @@ import { Container, Checkmark } from './style';
 
 interface CheckboxProps {
     children?: ReactNode;
+    onChange?: (event: React.FormEvent<HTMLInputElement>) => void;
     checked?: boolean;
     disabled?: boolean;
 }
 
 const Checkbox = ({
     children,
+    onChange,
     checked = false,
     disabled = false
 }: CheckboxProps): JSX.Element => {
     return (
         <Container disabled={disabled}>
             { children }
-            <input type="checkbox" defaultChecked={checked} disabled={disabled} />
+            <input 
+                type="checkbox" 
+                defaultChecked={checked} 
+                disabled={disabled} 
+                onChange={onChange} 
+            />
             <Checkmark className="checkmark" disabled={disabled} />
         </Container>
     );
--- a/src/components/Checkbox/index.tsx
+++ b/src/components/Checkbox/index.tsx
@@ -4,7 +4,7 @@ import { Container, Checkmark } from './style';
 
 interface CheckboxProps {
     children?: ReactNode;
-    onChange?: (event: React.FormEvent<HTMLInputElement>) => void;
+    onChange?: (checked: boolean) => void;
     checked?: boolean;
     disabled?: boolean;
 }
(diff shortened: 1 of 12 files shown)


src/components/Checkbox/index.tsx
```diff
@@ -4,18 +4,27 @@ import { Container, Checkmark } from './style';
...
-        <Container>
+        <Container disabled={disabled}>
             { children }
-            <input type="checkbox" defaultChecked={checked} />
-            <Checkmark className="checkmark" />
+            <input 
+                type="checkbox" 
+                defaultChecked={checked} 
+                disabled={disabled} 
+                onChange={onChange}
```
Cnordbo: This could potentially cause an invalid reference if the checkbox is re-rendered between onChange and when the receiver actually received the event (input element might be replaced with a new one) - Maybe simplify enough to just return a true/false value?

src/core/services/NotificationService/index.tsx
```diff
@@ -26,10 +29,11 @@ const Notification = (props: NotificationProps): JSX.Element => {
  */
 export const showSnackbarNotification = (
     message: string,
-    duration: number
+    duration: number,
```
Cnordbo: Could you add a default value to duration while in here? :) 5000 seems to be the one thats used troughout the application.

src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -468,6 +480,25 @@ class PreservationApiClient extends ApiClient {
         }
     }
 
+    async recordTagRequirementValues(tagId: number | null, recordValues: TagRequirementRecordValues, setRequestCanceller?: RequestCanceler): Promise<void> {
```
Cnordbo: tagId can never be set to null.

src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/Requirements.tsx
```diff
@@ -1,21 +1,110 @@
...
+
+        // has unsaved changes
+        if (requirementValues.findIndex(requirement => requirement.requirementId == requirementId) > -1) {
+            return false;
+        }
+
+        return isReadyToBePreserved;
+    };
+
+    const getNumberField = (requirementId: number, field: TagRequirementField): JSX.Element => {
```
Cnordbo: These can probably be separated out to different field components.

src/components/Checkbox/index.tsx
kfaln: Function accessing the event target moved to the component itself.

src/core/services/NotificationService/index.tsx
kfaln: Added 5000 default.

src/modules/Preservation/http/PreservationApiClient.ts
kfaln: Done.

src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/Requirements.tsx
kfaln: Moved to separate components.

PR #39, opened 2020-02-25
Preserve all requirements for a tag

kfaln:
* Tag flyout - header: "preserved this week" button.
* Refresh requirements after preserve tag.
* Refresh tag details after preserve requirement.
* UI tweaks.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
@@ -9,38 +9,41 @@ import { usePreservationContext } from '../../../../context/PreservationContext'
 import { showSnackbarNotification } from './../../../../../../core/services/NotificationService';
 
 interface PreservationTabProps {
-    tagId: number | null;
-    tagDetails: TagDetails | undefined;
-    getTagDetails: () => void;
-    updateRequirements: number;
+    tagDetails: TagDetails | null;
+    refreshTagDetails: () => void;
 }
 
 const PreservationTab = ({
-    tagId,
     tagDetails,
-    getTagDetails,
-    updateRequirements
+    refreshTagDetails
 }: PreservationTabProps): JSX.Element => {
-    const [tagRequirements, setTagRequirements] = useState<TagRequirement[]>();
+    const [tagRequirements, setTagRequirements] = useState<TagRequirement[] | null>(null);
     const { apiClient } = usePreservationContext();
 
-    const getTagRequirements = async (id: number): Promise<void> => {
-        try {            
-            const tagRequirements = await apiClient.getTagRequirements(id);
-            setTagRequirements(tagRequirements);
-        }
-        catch (error) {
-            console.error(`Get TagRequirements failed: ${error.message}`);
-            showSnackbarNotification(error.message, 5000, true);
+    const getTagRequirements = async (): Promise<void> => {
+        if (tagDetails) {
+            try {            
+                const tagRequirements = await apiClient.getTagRequirements(tagDetails.id);
+                setTagRequirements(tagRequirements);
+            }
+            catch (error) {
+                console.error(`Get TagRequirements failed: ${error.message}`);
+                showSnackbarNotification(error.message, 5000, true);
+            }
         }
     };
 
+    useEffect(() => {
+        setTagRequirements(null); // force full refresh
+        getTagRequirements();
+    }, [tagDetails]);    
+
     const recordTagRequirementValues = async (values: TagRequirementRecordValues): Promise<void> => {
-        if (tagId !== null) {
+        if (tagDetails) {
             try {
-                setTagRequirements(undefined); // trigger the spinner
-    
-                await apiClient.recordTagRequirementValues(tagId, values);            
+                setTagRequirements(null); // trigger the spinner
+        
+                await apiClient.recordTagRequirementValues(tagDetails.id, values);            
                 showSnackbarNotification('Requirement values saved', 4000, true);
             }
             catch (error) {
@@ -48,30 +51,20 @@ const PreservationTab = ({
                 showSnackbarNotification(error.message, 6000, true);
             }
             finally {
-                getTagDetails();
-                getTagRequirements(tagId);                
+                refreshTagDetails(); // will also trigger refresh of requirements           
             }
         }
     };
 
     const isReadOnly = (): boolean => {
-        return tagDetails 
-            ? tagDetails.status.toLowerCase() !== 'active' 
-            : false;
+        if (tagDetails) {
+            tagDetails.status.toLowerCase() !== 'active';
+        }
+
+        return false;
     };
 
-    useEffect(() => {
-        if (tagId !== null) {
-            // re-render requirements when "header" actions are executed
-            if (updateRequirements > 0) {
-                setTagRequirements(undefined);
-            }
-
-            getTagRequirements(tagId);
-        }
-    }, [tagId, updateRequirements]);    
-
-    if (tagDetails === undefined) {
+    if (tagDetails === null) {
         return <div style={{margin: 'calc(var(--grid-unit) * 5) auto'}}><Spinner medium /></div>;
     }
 
@@ -111,7 +104,6 @@ const PreservationTab = ({
                 requirements={tagRequirements} 
                 readonly={isReadOnly()} 
                 recordTagRequirementValues={recordTagRequirementValues} 
-                updateRequirements={updateRequirements}
             />
         </Container>
     );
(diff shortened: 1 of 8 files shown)


src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
```diff
@@ -57,9 +62,14 @@ const PreservationTab = ({
 
     useEffect(() => {
         if (tagId !== null) {
+            // re-render requirements when "header" actions are executed
+            if (updateRequirements > 0) {
+                setTagRequirements(undefined);
```
Cnordbo: use `null` if possible. `undefined` is basically "this has never existed"

src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
```diff
@@ -101,6 +111,7 @@ const PreservationTab = ({
                 requirements={tagRequirements} 
                 readonly={isReadOnly()} 
                 recordTagRequirementValues={recordTagRequirementValues} 
+                updateRequirements={updateRequirements}
```
Cnordbo: this sounds like an action, do we have a better way of defining properties and functions that will help us be consistent and make it easier to read now and in the future?

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -79,17 +114,28 @@ const TagFlyout = ({
...
+                        )}
+                        {!isLoadingPreserveTag && ('Preserved this week')}
+                    </Button>                    
                     <Button variant='ghost' title='Close' onClick={close}>
                         <CloseIcon />
                     </Button>                            
                 </HeaderActions>                        
             </Header>
+            <StatusLabel status={tagDetails && tagDetails.status}>
+                <span style={{marginLeft: 'var(--grid-unit)', marginRight: 'var(--grid-unit)'}}>
```
Cnordbo: `margin: '0 var(--grid-unit)'` covers both left and right.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -20,29 +21,55 @@ const TagFlyout = ({
...
+            catch (error) {
+                console.error(`Get TagDetails failed: ${error.message}`);
+                showSnackbarNotification(error.message, 5000, true);
+            }
         }
     };
 
-    useEffect(() => {
+    const preserveTag = async (): Promise<void> => {
         if (tagId !== null) {
```
Cnordbo: Is this module suppose to work without a tagId? if not, tagId should be non-optional - helps avoid the null checking everywhere.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -20,29 +21,55 @@ const TagFlyout = ({
...
-            getTagDetails(tagId);
+            try {
+                setIsLoadingPreserveTag(true);
```
Cnordbo: could `isPreserveringTag` be a more explanatory state here?

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -20,29 +21,55 @@ const TagFlyout = ({
...
+                await apiClient.preserveSingleTag(tagId);
+                showSnackbarNotification('Tag has been preserved for this week.', 5000, true);
+            }
+            catch (error) {
+                console.error(`Preserve tag failed: ${error.message}`);
+                showSnackbarNotification(error.message, 5000, true);
+            }
+            finally {
+                setIsLoadingPreserveTag(false);
+                setUpdateRequirements(updateRequirements + 1);
```
Cnordbo: when using existing state, best practices is to use functions when setting state. 
`setUpdateRequirements(existingRequirements => existingRequirements + 1)`

Cnordbo: What about just triggering a new call to getTagDetails? The new object is then automatically passed down to the PreservationTab, which in turn can reset its `updateRequirements` based on which property is changed.

This will also help us with keeping the view in the correct state as server side.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -20,29 +21,55 @@ const TagFlyout = ({
...
     }, [tagId]);
 
     const getTabContent = (): JSX.Element => {
         switch (activeTab) {
             case 'preservation':
-                return <PreservationTab tagId={tagId} tagDetails={tagDetails} />;
+                return <PreservationTab 
+                    tagId={tagId} 
+                    tagDetails={tagDetails} 
+                    getTagDetails={getTagDetails}
```
Cnordbo: Could `refreshTagDetails` as prop name be a better name here, as getTagDetails doesnt actually return anything?

src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
kfaln: Changed to null where possible.

src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
kfaln: This was a bit of an odd case with a mixed flavor - a property triggering an action. It's now refactored away, anyway, so nevermind :)

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: True dat.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: Short answer: No.
It was actually your idea to allow null instead of the original solution which was to set the initial value to 0 (we need a default value, otherwise TS gives us Undefined). Your argument was to set it to Null rather than 0 to indicate that the value was uninitialized.

I don't like the null checks either, so it's back to non-optional with 0 as default value now.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: Yes.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: Yes.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: I tried this initially and was having trouble getting the requirements to update. That's why the requirements are set to null (previously undefined) before getting them again, to force an update. It works ok now by calling getTagDetails and checking for updates on that to trigger the rest. The `updateRequirements` stuff is removed.

kfaln:
@Cnordbo done the needful as discussed (tagDetails vs null checking). It works fine.

src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
```diff
@@ -4,66 +4,88 @@ import { Container, TagDetailsContainer, Details, GridFirstRow, GridSecondRow, R
...
 }: PreservationTabProps): JSX.Element => {
-    const [tagRequirements, setTagRequirements] = useState<TagRequirement[]>();
+    const [tagRequirements, setTagRequirements] = useState<TagRequirement[] | null>(null);
     const { apiClient } = usePreservationContext();
 
-    const getTagRequirements = async (id: number): Promise<void> => {
+    const getTagRequirements = async (): Promise<void> => {
         try {            
-            const tagRequirements = await apiClient.getTagRequirements(id);
+            const tagRequirements = await apiClient.getTagRequirements(tagDetails.id);
```
Cnordbo: Variable naming: 
Outer scope already contains `tagRequirements`

PR #41, opened 2020-02-26
Start preservation on a tag

kfaln:
"Play" button in tag details header.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
@@ -88,7 +88,7 @@ const TagFlyout = ({
         return tagDetails.readyToBePreserved;
     };
 
-    const getTagStatus = (): PreservationStatus => {
+    const getTagPreservationStatus = (): PreservationStatus => {
         if (tagDetails)
         {
             switch (tagDetails.status.toLowerCase()) {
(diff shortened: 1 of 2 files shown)


src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -3,13 +3,21 @@ import React, { useState, useEffect } from 'react';
...
 import { TagDetails } from './types';
 import Spinner from '../../../../../components/Spinner';
 
+enum PreservationStatus {
+    NotStarted,
+    Active,
+    Completed,
+    Unknown
+}
+
```
Cnordbo: Try assigning the string equivalent? 

ex:
```js
enum PreservationStatus {
    NotStarted = 'notstarted',
    Active = 'active',
    Completed,
    Unknown
}
```

This allows us to do the comparison directly where needed. `tagDetails.status.toLowerCase() === PreservationStatus.NotStarted` ( theory )

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -55,6 +64,49 @@ const TagFlyout = ({
...
+                case 'completed':
+                    return PreservationStatus.Completed;                
+            }
+        }
+
+        return PreservationStatus.Unknown;
+    };
+
+    const preservationIsNotStarted = (): boolean => getTagPreservationStatus() === PreservationStatus.NotStarted;
+    const preservationIsStarted = (): boolean => getTagPreservationStatus() === PreservationStatus.Active;
```
Cnordbo: These can be set as variables instead of using them as methods. (every re-render will re-calculate them)

ex: 
```js
const preservationIsNotStarted = getTagPreservationStatus() === PreservationStatus.NotStarted;
```

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: It works.

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
kfaln: Done.

PR #49, opened 2020-03-05
Modal dialog service

arildbjerkenes:


src/core/services/ModalDialogService/style.ts
```diff
@@ -1,35 +1,50 @@
...
-    background-color: white;
-    padding: calc(var(--grid-unit) * 2);
-    width: calc(var(--grid-unit) * 100);
-    margin-top: 300px;    
-    padding-bottom: calc(var(--grid-unit) * 8);
-    `;
+export const DialogContainer = styled.div<{ width: number }>`
+    ${(props): any => `
+        display: block;
+        width: ${props.width}px;
```
kfaln: I'm not a big fan of pixel-based sizes. I think it can cause problems when viewing the solution on other screen types (not "responsive"). I would rather prefer using a relative size in percentage. We could however prepare for this by letting the size property be a string, instead of number (and not hard-coding 'px' at the end). That way each developer can determine what's best for each context, by for instance sending in `'600px'` or `'50%'`

src/core/services/ModalDialogService/style.ts
```diff
@@ -1,35 +1,50 @@
...
+        background-color: white;
```
kfaln: Use tokens wherever possible: `tokens.colors.ui.background__default.rgba`

src/core/services/ModalDialogService/style.ts
```diff
@@ -1,35 +1,50 @@
...
+        box-shadow: 0px 3px 5px rgba(0, 0, 0, 0.2), 0px 1px 18px rgba(0, 0, 0, 0.12),
```
kfaln: Check out `elevation.js` in eds-tokens. There is one called 'above_scrim' that might be usable here...

src/core/services/ModalDialogService/style.ts
```diff
@@ -1,35 +1,50 @@
...
+`;
+
+export const Title = styled.div`
+    padding-top: calc( var(--grid-unit) * 2 - 4px );
+    padding-right: calc(var(--grid-unit) * 2);
+    padding-left: calc(var(--grid-unit) * 2);
+`;
+
+export const Divider = styled.div`
+    border-top: 1px solid #DCDCDC;
```
kfaln: Try to find the color in tokens...

src/core/services/ModalDialogService/index.tsx
```diff
@@ -26,16 +27,20 @@ const ModalDialog = (props: ModalDialogProps): JSX.Element => {
         render(<></>, modalDialogContainer);
```
kfaln: Perhaps hide the dialog **before** running the callback? Not sure how this works in real life with async and stuff, but if it's a "big" callback, it might need to finish before the dialog is hidden. Imagine that the callback renders a new dialog :)

src/core/services/ModalDialogService/index.tsx
```diff
@@ -26,16 +27,20 @@ const ModalDialog = (props: ModalDialogProps): JSX.Element => {
...
-            <ModalContent>
-                {props.message}
+        <Scrim>
+            <DialogContainer width={width}>
+                {props.title &&
+                    <Title>{props.title}</Title>
+                }
+                {props.title &&
+                    <Divider />
+                }
```
kfaln: Wrap these in a `<Header>` (or similar) and remove `<Divider>`.

src/core/services/ModalDialogService/index.tsx
```diff
@@ -26,16 +27,20 @@ const ModalDialog = (props: ModalDialogProps): JSX.Element => {
...
+                {props.content &&
+                    <Content>{props.content}</Content>
+                }
                 {props.confirmText &&
                     <ButtonContainer>
                         <Button title='Confirm' onClick={confirm}>
```
kfaln: Not sure about the hard-coded titles on confirm and cancel buttons. The tooltip might be inconsistent with the button text/meaning, since we are allowing anything to be displayed on the buttons. Perhaps best to remove title for now? The button text should be enough...

src/core/services/ModalDialogService/style.ts
arildbjerkenes: Forgot our discussion...

arildbjerkenes: I will support percentage (string), but will use px for my transfer dialog. I will let the content decide the width, not the user's screen.

src/core/services/ModalDialogService/index.tsx
arildbjerkenes: Agree

src/core/services/ModalDialogService/index.tsx
arildbjerkenes: Can't do. If we follow the design, the padding for divider is different from title.

src/core/services/ModalDialogService/index.tsx
arildbjerkenes: Forgot to remove the title.

src/core/services/ModalDialogService/index.tsx
kfaln: Then at least wrap <Title> and <Divider> in a <div> so that you don't need the two conditional renderings.

src/core/services/ModalDialogService/index.tsx
arildbjerkenes: Have generalized the naming on the buttons.

src/core/services/ModalDialogService/index.tsx
```diff
@@ -1,72 +1,90 @@
...
 export const showModalDialog = (
-    message: string,
-    cancelText?: string,
-    confirmText?: string,
-    confirmCallback?: any,
+    title: string,
+    content?: ReactNode,
+    width?: string,
+    buttonOneText?: string,
+    buttonOneCallback?: any,
```
kfaln: Don't use `any`. You could type it `() => void | null`

src/core/services/ModalDialogService/index.tsx
```diff
@@ -1,72 +1,90 @@
...
+ * @param buttonTwoText   Text on the second button. 
+ * @param buttonTwoCallback  Callback function that will be called when clicking on the second button. 
  */
 export const showModalDialog = (
-    message: string,
-    cancelText?: string,
-    confirmText?: string,
-    confirmCallback?: any,
+    title: string,
+    content?: ReactNode,
```
kfaln: I don't like using optional `?` parameters. Typescript will type them as for instance `string | undefined`. We should avoid undefined. I would probably type them for instance `string | null` instead of making them optional. This will also make any null-checks on props more clear.

src/core/services/ModalDialogService/index.tsx
```diff
@@ -1,72 +1,90 @@
...
+                    <Title>{props.title}</Title>
                 }
-                {props.cancelText &&
+                {props.title &&
+                    <Divider />
+                }
+                {props.content &&
+                    <Content>{props.content}</Content>
+                }
+                {props.buttonTwoText &&
```
kfaln: Something strange going on here.. why is button two coming before button one?
I'm thinking there should be a common `<ButtonContainer>` where all buttons live...

Psudo code:
```
<ButtonContainer>
if (button 1) && <Button ...>
if (button 2) && <Button ...>
</ButtonContainer>
```

src/core/services/ModalDialogService/index.tsx
```diff
@@ -1,72 +1,90 @@
...
-    const confirm = (): void => {
-        props.confirmCallback();
+    const buttonTwoHandler = (): void => {
         render(<></>, modalDialogContainer);
-    };
-
-    window.onclick = (e: any): void => {
-        if (e.target.id === 'Modal') {
-            cancel();
+        if (props.buttonTwoCallback != null) {
```
kfaln: If prop was typed as `() => void | null` -- you could do `if (props.buttonTwoCallback) { ... }`

src/core/services/ModalDialogService/index.tsx
Cnordbo: I dont agree, and think optional parameters is something we should use. Null checks or undefined checks would be the same, as both are falsy values. so you only check `if (confirmCallback)` and not `if (confirmCallback === null)` the result would be the same for a nullable property.
a nullable property is also specificed as undefined if not used. 

What i do agree with, is the use of null when doing standard functional calls. 
For instance, `dialogService.showDialog("hello", null, "OK")` remember that we also got support for default values here, so that we leave out the remaining parameters. 

Ex. `width: string | null = null`

kfaln: Ok, but when using optional parameters, should you at least type it e.g `string | null` ?
I guess this will force typescript to not type it potentially undefined, which is something you have previously stated is something we should avoid...

Cnordbo: My statement has previosly been that you never define a type as undefined, if i remember correctly. `undefined` is not a type, and indicates the "lack of property" vs the "property doesnt have a value" 
If i remember the previous case correctly: 
So when using `const [..] = setState<string | undefined>(undefined)` i would endorse the use of `null` over `undefined`. 

But since this is parameters / properties used by "external" usage, i would much rather write `<MyDialogService />` instead of `<MyDialogService title={null} />` which would be required when using `string | null` on a component prop.

arildbjerkenes: If i use '() => void' instead of 'any' I'm not allowed to use null when calling the function.  I have used ' | null' when I have parameters in the start of the parameter list, and then ? when they are in the end of the list, so that I don't need null, null, null in the end.

src/core/services/ModalDialogService/index.tsx
arildbjerkenes: No so strange, but I can make it simpler. Just for you.... :-).

PR #50, opened 2020-03-10
Feature/71675 ui improvements on transfer dialog

arildbjerkenes:


src/core/services/ModalDialogService/index.tsx
```diff
@@ -1,72 +1,93 @@
...
+    buttonOneCallback: (() => void) | null;
+    buttonTwoText: string | null;
+    buttonTwoCallback: (() => void) | null;
 }
 
 const ModalDialog = (props: ModalDialogProps): JSX.Element => {
 
-    const cancel = (): void => {
+    const buttonOneHandler = (): void => {
         render(<></>, modalDialogContainer);
```
Cnordbo: This should not be handled by calling render  - but rather trough a visibility state.
Same thing goes for buttontwo. You never call render yourself within a component.

src/modules/Preservation/views/ScopeOverview/RequirementIcons.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+
+import { PreservedTag, Requirement } from './types';
+import { RequirementsContainer, RequirementIcon } from './ScopeTable.style';
```
kfaln: Create RequirementIcons.style.ts and move related styling and logic.

Cnordbo:
I dont agree with the general approach with using something called a DialogService for displaying modals. 
Rather, it should be handled by the component that spawns it, where one can have a common `<Modal>` Styled component that can be used, as this is really all you need.

This would then allow of full customizability by each module that uses it, which is probably what is needed in the end anyway. with regards of size, buttons, closing behavior, styling, responsivenes ++.

src/modules/Preservation/views/ScopeOverview/TransferDialog.tsx
```diff
@@ -3,97 +3,86 @@ import { PreservedTag } from './types';
 import { Typography } from '@equinor/eds-core-react';
 import { tokens } from '@equinor/eds-tokens';
 import Table from './../../../../components/Table';
-import { getRequirementColumn, isTagOverdue } from './ScopeTable';
+import RequirementIcons from './RequirementIcons';
 import { Toolbar } from './ScopeTable.style';
```
kfaln: Create TransferDialog.style.ts and establish relevant styling (copy/paste if you need to).

arildbjerkenes:
Agreed with Christer to create a new PBI to change the overall implementation on the dialog component.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -90,13 +103,30 @@ const ScopeOverview: React.FC = (): JSX.Element => {
...
-            showModalDialog(`${selectedTags.length} selected tags. Please confirm to transfer all selected tags, or go back to list.`, 'Back to list', 'Transfer', transfer);
-        } else {
-            showModalDialog(`${numTagsNotTransferable} tag(s) are not transferable.`, 'Back to list');
-        }
+        const transferableTags: PreservedTag[] = [];
+        const nonTransferableTags: PreservedTag[] = [];
+
+        //Tag-objects must be cloned to avoid issues with data in scope table 
+        selectedTags.map((tag) => {
+            const newTag: PreservedTag = JSON.parse(JSON.stringify(tag));
```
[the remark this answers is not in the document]
arildbjerkenes: Ok, shallow copy is good enough here.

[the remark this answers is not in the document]
arildbjerkenes: No wait, the PreservedTag includes an array with Requirement objects. Guess spread will not handle this.

Cnordbo: Is this really needed? Does the tags get manipulated somewhere else?  Cant you just sort them without copying? 

You can also do 
```js
const transferableTags: PreservedTag[] = selectedTags.filter((tag) => tag.readyToBeTransferred)
const nonTransferableTags: PreservedTag[] = selectedTags.filter((tag) => !tag.readyToBeTransferred)
```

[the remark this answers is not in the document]
arildbjerkenes: It works with spread here, because we don't need a proper clone of the requirement objects. However, if we need deep clone, i guess the JSON way of doing it is the preferred?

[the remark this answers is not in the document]
Cnordbo: Need to look into this if thats the case. JSON way is actually really slow, so other options might be better suited. (recursive re-assignment or something) - but i havent looked into it, as the need for doing so is not common. Try to re-consider if this is needed. I cant see the reason why it needs to be copied (as per my other comment).

src/core/services/ModalDialogService/index.tsx
arildbjerkenes: I include this in a new PBI.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
arildbjerkenes: The code you show here is exactly how I had it in the first place. However, it gave some strange behavior when returning to the scope list. Material-table adds variables to the objects. This get messed up when I use the same objects in a new table in the dialog.

PR #54, opened 2020-03-17
TreeView component

kfaln:
This should be considered the first version of a work in progress.

The component needs some modification when the first (detail) view is to be implemented in the new Library, mainly the following.
- Async handling when fetching children for nodes.
- View routing/rendering when clicking on nodes (which view to display).

The TreeView currently handles static data, has working expand/collapse logic and styling.


src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+        });
+
+        const newTreeData = [...treeData];
+        newTreeData.splice(collapsingNodeIndex + 1, childCount);
+
+        setTreeData(newTreeData);
+    };
+
+    const expandNode = (node: NodeData): void => {
+        if (!node.getChildren || node.isExpanded) {
```
Cnordbo: is this ment to be `!node.hasChildren` ?

src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+ * @param id Unique node identifier across all nodes (number or string).
+ * @param name Node display name.
+ * @param hasChildren Indicates node has children and can be expanded.
+ * @param getChildren Function to load children.
+ * @param onClick Function to load details view.
+ * @param parentId Internal: reference to parent node (handled by component).
+ * @param isExpanded Internal: indicates node is currently expanded (handled by component).
+ * @param children Internal: reference to child nodes (handled by component).
+ */
+export interface NodeData {
```
Cnordbo: Is this an un-needed export?

src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    id: number | string;
+    name: string;
+    hasChildren: boolean;
+    getChildren: (() => NodeData[]) | null;
```
Cnordbo: getChildren could probably always return a empty list, and should always be configured on the nodes.

src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    onClick: (() => void) | null;
```
Cnordbo: Optional property? 
`onClick?: (() => void)`

src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    const getNodeLink = (node: NodeData): JSX.Element => {
+        const hasOnClickEvent = node.onClick;
+
+        return (
+            <NodeName 
+                hasChildren={node.hasChildren} 
+                isExpanded={node.isExpanded}>
+                {
+                    hasOnClickEvent && (
```
Cnordbo: `node.onClick && ...`
No need for the extra variable,

src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+                        <NodeLink
+                            isExpanded={node.isExpanded} 
+                            onClick={(): void => {
```
Cnordbo: Existence checking is already done before loading this code. 
Can be replaced with: `onClick={node.onClick}`

src/components/TreeView/index.tsx
kfaln: `hasChildren` was just a convenience prop, basically meaning the same as `getChildren` being defined/not null. `hasChildren` is removed/replaced with just checking `getChildren` (also moved elsewhere).

src/components/TreeView/index.tsx
kfaln: Parent components will need to send in data as a prop. The idea is that `NodeData` will be the data interface. However, it contains some props the parent won't need, so it's now split in two and only one of them is exported (`TreeViewNode`).

src/components/TreeView/index.tsx
kfaln: The parent container will need to supply functions for fetching children, if there are any. It's now changed to optional.

src/components/TreeView/index.tsx
kfaln: Yes.

src/components/TreeView/index.tsx
kfaln: Yes.

src/components/TreeView/index.tsx
kfaln: Yes.

PR #55, opened 2020-03-19
Feature/71669 support paging for preservation scope

arildbjerkenes:
The solution here might seem a bit strange. Because the data viewed by material-table is not given by a state property, we don't get a rerender of the table unless material-table believe it is necessary. To trigger this, i got the recommendation to use a reference to the table, and call onQueryChange. 

I have also added sorting here. 

I have temporary commented out the isLoading feature. It does not work properly, when we use remote data for the scope list. I will add a new PBI to figure out this.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
@@ -8,8 +8,8 @@ import RequirementIcons from './RequirementIcons';
 import { isTagOverdue, getFirstUpcomingRequirement } from './ScopeOverview';
 
 interface ScopeTableProps {
-    getTags: (page: number, pageSize: number) => Promise<PreservedTags | null>;
-    isLoading: boolean;
+    getTags: (page: number, pageSize: number, orderBy: string | null, orderDirection: string | null) => Promise<PreservedTags | null>;
+    //isLoading: boolean;
     setSelectedTags: (tags: PreservedTag[]) => void;
     showTagDetails: (tag: PreservedTag) => void;
     setRefreshScopeListCallback: (callback: () => void) => void;
@@ -17,7 +17,7 @@ interface ScopeTableProps {
 
 const ScopeTable = ({
     getTags,
-    isLoading,
+    //isLoading,
     setSelectedTags,
     showTagDetails,
     setRefreshScopeListCallback,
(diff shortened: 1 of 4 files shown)


src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
```diff
@@ -1,27 +1,38 @@
...
 }: ScopeTableProps): JSX.Element => {
 
+    const ref = React.createRef();
+
+    const refreshScopeList = (): void => {
+        const referanse: any = ref.current;
+        referanse?.onQueryChange();
+    };
+
+    setRefreshScopeListCallback(refreshScopeList);
```
kfaln: Might be able to simplify this a bit:

``` js
    setRefreshScopeListCallback(() => {
        if (ref.current) {
            (ref as any).onQueryChange();
        }
    });
```

src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
```diff
@@ -58,26 +69,56 @@ const ScopeTable = ({
...
+                                totalCount: result.maxAvailable
+                            }) : null;
+                    });
+                })
+            }
             options={{
                 showTitle: false,
                 draggable: false,
                 selection: true,
                 pageSize: 10,
```
kfaln: Should we increase the default pagesize to, say.. 50 ?

kfaln:
Weird bug:
1. Change pagesize to 50.
2. Click on a tag to open details.
3. Close the flyout. Notice that pagesize has popped back to 10 (or default size).
4. Going to next page now fails (major crash, see console).

src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
arildbjerkenes: Let do it.

src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
arildbjerkenes: Agree

arildbjerkenes:
> 2\. t

When coming from tagflyout, we will now stay at the page size you have. However, the page size will be reset after creating a new tag, or clicking on the 'Preservation' tab. I suppose we later will store the page size as user preference, so currently I think we should leave it like this. 

Note: Things work, but there are some error messages in the console, when opening tag-details. The first time we enter the tag-details, after restarting, we get a strange error message from the Typography component. When we open tag-details several times, we get an error message indicating that something is wrong. Still, things work. I suggest that we create a PBI to look into this. We might need Christer to help us out on this.

kfaln:
> Note: Things work, but there are some error messages in the console, when opening tag-details. The first time we enter the tag-details, after restarting, we get a strange error message from the Typography component. When we open tag-details several times, we get an error message indicating that something is wrong. Still, things work. I suggest that we create a PBI to look into this. We might need Christer to help us out on this.

I have also noticed this error message, but have not invested the time to look more into it. Definitely something to do with an eds-component, perhaps an issue with it being initialized before having data to display. I'm only getting it the first time opening tag details. Subsequent tag details don't produce errors (no matter how many times), and it works without problem in all cases. Agree we should try to eliminate it at some point.

PR #56, opened 2020-03-27
Feature/71466 ui new library module in preservation framework

arildbjerkenes:
At this point we should not try to implement the complete library framework for library. There are a lot of things to consider and discuss. For now we implement enough to be able to support what we need for Preservation. 

The treeview component is not complete at this point. We should include that a node is shown as 'selected', and the component should expose the path to the selected node (e.g. array with node-names). Maybe we should store an object representing the node (the object we fetch from server). 

There are some 'prof of concept' code here as well. 

Currently the node selected is not reflected in the url. My plan was that instead of changing a state variable, we should do a routing, when a node is selected. I suggest we add a new PBI to add this.


src/modules/PlantConfig/style.ts
```diff
@@ -0,0 +1,24 @@
...
+    font-family: Equinor;
+    font-style: normal;
+    color: var(--text--primary-white);
+    font-size: 14px;
+    padding: 10px 16px;
+    border-radius: 4px;
+    border: none;
+    background-color: ${tokens.colors.interactive.primary__resting.rgba};
+    font-weight: 500;
+`;
```
kfaln: This is probably not used, here and also in `Preservation/style.ts`

src/modules/PlantConfig/views/Library/Library.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useEffect, useState } from 'react';
+//import withAccessControl from '../../../../core/security/withAccessControl';
```
kfaln: Remove

src/modules/PlantConfig/views/Library/LibraryTreeview/LibraryTreeview.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Container } from './LibraryTreeview.style';
+
+type LibraryTreeviewProps = {
+    setSelectedLibraryType: (libraryType: string) => void;
+    setSelectedLibraryItem: (libraryItem: string) => void;
+};
+
+const LibraryTreeview = (props: LibraryTreeviewProps): JSX.Element => {
+
+    interface Mode {
```
kfaln: Unused

src/modules/PlantConfig/style.ts
arildbjerkenes: Done

src/modules/PlantConfig/views/Library/Library.tsx
arildbjerkenes: done

src/modules/PlantConfig/views/Library/LibraryTreeview/LibraryTreeview.tsx
arildbjerkenes: done

PR #58, opened 2020-03-31
fix issue with update of scope table

arildbjerkenes:


src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
```diff
@@ -31,7 +31,7 @@ const ScopeTable = ({
 
     setRefreshScopeListCallback(() => {
         if (ref.current) {
-            (ref as any).onQueryChange();
+            (ref as any).current.onQueryChange();
```
Cnordbo: STOP using any!

arildbjerkenes: Fixed. Had to use any here: const ref: RefObject<any> = React.createRef();

Was not able to use MaterialTable at type here.

PR #59, opened 2020-04-03
Only post tags that can be transferred. 

arildbjerkenes:
I have included some other smaller fixes as well (only changed two code lines). 
- Product Backlog Item 72574: UI - Remove sorting on requirement type
- Product Backlog Item 72549: UI - TagNo misses area in sidepanel

Also, I added some new table size options on scope list.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
@@ -88,8 +88,8 @@ const ScopeOverview: React.FC = (): JSX.Element => {
         return Promise.resolve();
     };
 
-    const transferableTags: PreservedTag[] = [];
-    const nonTransferableTags: PreservedTag[] = [];
+    let transferableTags: PreservedTag[];
+    let nonTransferableTags: PreservedTag[];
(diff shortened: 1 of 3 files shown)


src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -88,12 +88,15 @@ const ScopeOverview: React.FC = (): JSX.Element => {
         return Promise.resolve();
     };
 
+    const transferableTags: PreservedTag[] = [];
+    const nonTransferableTags: PreservedTag[] = [];
```
kfaln, line 92: Will these arrays need to be reset, the same way as `selectedTags`? What happens if the user closes/cancels the transfer dialog, makes a different tag selection, and then opens a new dialog? And what about when the transfer is completed?

arildbjerkenes: This works because there will be a rerender each time a tag is selected/deselected (arrays are reset every time). I agree that we should not rely on the rerender, so I have made a change. We don't need a state variable for this.

PR #61, opened 2020-04-08
Record explicit requirement values for checkbox and number types

kfaln:


src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/RequirementCheckboxField.tsx
```diff
@@ -25,7 +25,7 @@ const RequirementCheckBoxField = ({
             checked={isChecked} 
             disabled={readonly}
             onChange={(checked: boolean): void => {
-                setFieldValue(requirementId, field.id, checked.toString());
+                setFieldValue(requirementId, field.id, checked);
```
Cnordbo, line 28: We should decide on a common naming strategy for props. 

`setFieldValue` looks like a logical name for an operation that updates local state. 
But when using props, you basically "dont know" whats happening at the parent level.
So this should be reflected in the naming that is used by the prop. 

Example:
`onFieldUpdate`, `onFieldValueChange`, `onTextChange`, `onInputChange`, `onChange` etc. 

The same way as native React components and HTML uses its properties. (onClick, onChange etc)

kfaln: I concur. Makes sense. I've updated the prop names to OnFieldChange in this case.

PR #73, opened 2020-04-23
72372   validate tag number

bratli:
Check to see if tag number generated is valid. Error message if not valid or if suffix has space.

Next is only enabled if all mandatory fields are filled in and it creates a valid tag no.

tests.


src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
```diff
@@ -154,67 +166,123 @@ const CreateAreaTag = (props: CreateAreaTagProps): JSX.Element => {
...
-                    style={{ maxWidth: '200px' }}
-                    label="Tag number suffix (space not allowed)"
+                    label="Tag number suffix"
                     inputRef={suffixInputRef}
                     placeholder="Write Here"
-                    helpertext="Text added to the end of the tagno."
-                    onChange={(e: any): void => props.setSuffix(e.target.value)}
+                    helperText="Spaces are not allowed"
+                    helperIcon={icon}
+                    variant={variant}
```
Cnordbo, line 283: can variant just be based on if there is an error message or not? instead of requiring its own state? 
`variant={tagNoErrorMessage ? 'error' : 'default'}`
or the icon itself? 
`variant={icon ? 'error' : 'default'}`

variant always seems to change with the icon being shown.

src/declarations/eds-core-react.d.ts
```diff
@@ -1 +1 @@
-declare module '@equinor/eds-core-react';
\ No newline at end of file
+declare module '@equinor/eds-core-react';
```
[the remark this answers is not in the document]
bratli: Will take note!

src/modules/Preservation/views/AddScope/CreateAreaTag/CreateAreaTag.tsx
bratli: smart

PR #74, opened 2020-04-23
Feature/72417   start   show dialog

bratli:
Enable start button when one or more tags are selected. On click open dialog that shows which or the selected tags can and cannot be started. Same as for transfer.


src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -135,15 +122,56 @@ const ScopeOverview: React.FC = (): JSX.Element => {
...
     };
 
+    let startableTags: PreservedTag[];
+    let nonStartableTags: PreservedTag[];
+
+    const startPreservation = async (): Promise<void> => {
+        try {
+            await apiClient.startPreservation(startableTags.map(t => t.id));
+            refreshScopeList();
+            setSelectedTags([]);
```
Cnordbo, line 132: Verify if these two should change order. (Just so that it doesnt refresh, select all the "old" tags, and then reset the selection)

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -18,6 +18,7 @@ import { showModalDialog } from '../../../../core/services/ModalDialogService';
 import { PreservedTag, Requirement, PreservedTags, TagListFilter } from './types';
 import ScopeTable from './ScopeTable';
 import TransferDialog from './TransferDialog';
+import StartDialog from './StartDialog';
```
Cnordbo, line 21: Naming: `StartPreservationDialog` ? As thats the only thing that can be started from this dialog? :)

PR #75, opened 2020-04-23
Feature/72416   bulkpreserve   dialog

bratli:
"Preserved this week" button enabled when one tag or more is selected. Dialoge is opened (same as transfer) with more info.


src/modules/Preservation/views/ScopeOverview/PreservedDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+                color: isTagOverdue(rowData) && tokens.colors.interactive.danger__text.rgba,
+            }),
+        }}
+        components={{
+            Toolbar: (): any => (
+                <Toolbar>
+                    <Typography style={{ color: toolbarColor }} variant='h6' >{numTags} {toolbarText}</Typography>
+                </Toolbar>
+            )
+        }}
```
Cnordbo, line 68: This is OK, but dont try to over-simplify things just because it repeats itself once. 
You have basically just done an extra abstraction of an already abstracted components settings here..

src/modules/Preservation/views/ScopeOverview/PreservedDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+    toolbarColor: string;
+}
+
+const getRequirementIcons = (tag: PreservedTag): JSX.Element => {
+    return (
+        <RequirementIcons tag={tag} />
+    );
+};
+
+const PreservedTable = ({
```
Cnordbo, line 27: Rule of thumb, 1 component per file. 

Otherwise it can be defined within functions of components.

src/modules/Preservation/views/ScopeOverview/__tests__/PreservedDialog.test.jsx
```diff
@@ -0,0 +1,57 @@
...
+        status: 'Active',
+        requirements: [{id: 1, requirementTypeCode: 'Grease'}],
+    }
+];
+
+describe('<PreservedDialog />', () => {
+
+    /** Because of API calls using effect hooks, we need to wrap everything in act */
+    it('Should only display nonpreservable tags when no preservable tags are selected', async () => {
+        await act(async () => {
```
Cnordbo, line 30: Wrapping tests in `act` is only needed when doing async calls that trigger state updates in `useEffect` normally.

PR #81, opened 2020-04-29
Fixed bug with missing data when autoscope.

arildbjerkenes:


arildbjerkenes:
Just a quick bug fix. Merging without review.

PR #82, opened 2020-04-30
Feature/71430   styling select requirements

bratli:
- Changed text from no items available to no items found
- Left aligned text in select-component
- Changed label text to match text field
- Background on hover item
- Correct padding on icons
- Optional on optional fields in Create Area Tag

### DIFF
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -146,7 +146,9 @@ const Select = ({
                 </ul>
             )}
             {isOpen && data.length <= 0 && !disabled && (
-                <li data-value={-1}>No items found</li>
+                <ul style={{boxShadow: 'none'}}>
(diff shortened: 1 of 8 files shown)


src/components/Select/index.tsx
```diff
@@ -147,9 +146,7 @@ const Select = ({
                 </ul>
             )}
             {isOpen && data.length <= 0 && !disabled && (
-                <ul>
-                    <li data-value={-1}>No items available</li>
-                </ul>
+                <li data-value={-1}>No items found</li>
```
Cnordbo, line 149: `li` has to have a `ul` parent to be valid HTML.

bratli: When everything is rendered it ends up being inside an ul (same as if items are found). Have added the `ul` back

PR #85, opened 2020-04-30
Scope Filter - performance boost when toggling filterpanel

Cnordbo:
## Switch from FC to Class Component for ScopeTable. 
As this PR stands, this causes the ScopeTable component to never update after first render. 
From initial testing, this seems to be fine, since all of our updates are triggered with `refreshScopeList` 

- but i need to look have a second look just so i dont miss anything. 
And it would be nice if you could take a look as well.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/ban-ts-ignore */
-import React, { RefObject } from 'react';
+import React, { RefObject, ReactNode } from 'react';
 import Table from './../../../../components/Table';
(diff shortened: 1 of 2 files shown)


src/modules/Preservation/views/ScopeOverview/ScopeTable.tsx
```diff
@@ -20,36 +20,40 @@ interface ScopeTableProps {
...
+
+    constructor(props: ScopeTableProps) {
+        super(props);
+        this.refObject = React.createRef();
+        this.getTagsByQuery = this.getTagsByQuery.bind(this);
+        this.getTagNoColumn = this.getTagNoColumn.bind(this);
+    }
+
+    shouldComponentUpdate(): boolean {
+        return false;
```
Cnordbo, line 34: This is basically the reason for the re-render.

PR #87, opened 2020-05-04
Add 'All' option to preservation status filter

Cnordbo:
Adds the 'All' option to action filter


Cnordbo:
Fonts changed in: #88

PR #90, opened 2020-05-04
Rename action status title in filter

bratli:
Rename action status titles


src/modules/Preservation/views/ScopeOverview/ScopeFilter/ScopeFilter.tsx
```diff
@@ -61,24 +61,24 @@ const PRESERVATION_STATUS = [{
 }];
 
 const ACTION_STATUS = [{
-    title: 'All',
+    title: 'All actions',
```
janid1967, line 64: This is not the correct phrase. Remember that the filter is for tags having actions or not. Not the action itself.
Just "All" is more correct. We should discuss together with Lotte.

PR #92, opened 2020-05-06
Feature/72222 optimistic concurrency

paal123:
Supporting optimistic concurrency.

Adding RowVersion reflecting change in backend.
Adding support for updated row version in response to enable consecutive put calls.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
@@ -105,7 +105,7 @@ const PreservationTab = ({
 
     const saveRemarkAndStorageArea = async (remarkString: string, storageAreaString: string): Promise<void> => {
         try {
-            await apiClient.setRemarkAndStorageArea(tagDetails.id, remarkString, storageAreaString);
+            await apiClient.setRemarkAndStorageArea(tagDetails.id, remarkString, storageAreaString, tagDetails.rowVersion);
         } catch (error) {
             console.error('Edit failed: ', error.messsage, error.data);
             showSnackbarNotification(error.message);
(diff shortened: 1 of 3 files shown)


bratli:
Det ser ut som noe går galt hvis man oppdaterer en tag mer enn en gang. Altså åpner en tag, redigerer remark, også redigerer remark igjen.

paal123:
> Det ser ut som noe går galt hvis man oppdaterer en tag mer enn en gang. Altså åpner en tag, redigerer remark, også redigerer remark igjen.

Dette skyles at det kommer en endring backend også.

src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/PreservationTab.tsx
```diff
@@ -105,7 +106,8 @@ const PreservationTab = ({
 
     const saveRemarkAndStorageArea = async (remarkString: string, storageAreaString: string): Promise<void> => {
         try {
-            await apiClient.setRemarkAndStorageArea(tagDetails.id, remarkString, storageAreaString);
+            const updateRowVersion = await apiClient.setRemarkAndStorageArea(tagDetails.id, remarkString, storageAreaString, rowVersion);
```
Cnordbo, line 109: update**d**RowVersion ?

rowVersion should probably also be set by using `useRef` instead of `useState`, as useRef does not trigger a re-render. (and rowVersion changes doesnt need to trigger re-renders)

When it comes to naming, we should be explicit about how we name the rowVersion so that we can identify which "form" value its attached to. 

This can be done by either setting a object as the ref: 
```js
const rowVersions = useRef({
remark: null,
storageArea: null
})
```
**usage:** 
```js
apiClient.set....(rowVersions.current.remark)
```
**update:** 
```js
rowVersions.current = {...rowVersions.current, remark: 'new-rowVersion'}
```

### --- or ----

Using multiple useRefs
```js
const rowVersionRemark = useRef('');
const rowVersionStorageArea = useRef('');
```
**usage:** 
```js
apiClient.set....(rowVersionRemark.current)
```

**update:** 
```js
rowVersionRemark.current = 'new-row-version';
```

paal123: Nice! Did the update now, and it is much cleaner. @Cnordbo

PR #95, opened 2020-05-07
Feature/73424   complete button

bratli:
Add readyToBeCompleted in tag type
Add complete button on scope overview
Add complete dialog
Add complete endpoint


Cnordbo:
Can / Should the logic for completing this be moved to the actual Complete Dialog? 

ScopeOverview can get really really really big if we keep on adding everything to that component. 

We need to separate component logic wherever we can, where it makes sense.

bratli:
> Can / Should the logic for completing this be moved to the actual Complete Dialog?
> 
> ScopeOverview can get really really really big if we keep on adding everything to that component.
> 
> We need to separate component logic wherever we can, where it makes sense.

I can try to move all of the dialog logic inside the actual dialog. May have a few hours today where it isnt worth starting a brand new task.

PR #105, opened 2020-05-26
Improve requirement selector

bratli:
Improve requirement selector by grouping options. Can add titles within a select

### DIFF
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -74,7 +74,7 @@ const Select = ({
                     key={index}
                     tabIndex={0}
                 >
-                    <TitleContent className={index > 0 ? 'borderTop' : ''} >
+                    <TitleContent borderTop={index > 0} >
                         {itm.text}
                     </TitleContent>
                 </TitleItem>;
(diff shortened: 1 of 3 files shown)


src/components/Select/index.tsx
```diff
@@ -67,8 +68,17 @@ const Select = ({
...
 
     const createNodesForItems = (items: SelectItem[]): JSX.Element[] => {
-
         return items.map((itm, index) => {
+            if(itm.title) {
+                return <TitleItem
+                    key={index}
+                    tabIndex={0}
+                >
+                    <TitleContent className={index > 0 ? 'borderTop' : ''} >
```
Cnordbo, line 77: does this actually work? i just tried something similar which didnt (it discarded the classname on styled-components)
you can use a standard prop to get the same result. 

index.tsx
```js
....
<TitleContent borderTop={index > 0}>
....
```
styles.ts: 
```js
$(props => props.borderTop && css`
border-top: ....
`)
```

bratli: This works, yes, but I will look into your methods as well. Are those more 'correct' than this solution?

Cnordbo: When using styled-components, i think this is the preferred way.

PR #111, opened 2020-05-28
Feature/73877   unchecking checkbox not working

bratli:
Checking/unchecking checkbox working as expected

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/Requirements.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/Requirements.tsx
@@ -170,11 +170,7 @@ const Requirements = ({
         if (requirement && field.currentValue) {
             const fieldIndex = requirement.checkBoxValues.findIndex(f => f.fieldId == field.id);
             if (fieldIndex > -1){
-                if (requirement.checkBoxValues[fieldIndex].isChecked) {
-                    return true;
-                } else {
-                    return false;
-                }
+                return requirement.checkBoxValues[fieldIndex].isChecked;
             }
         }
         return field.currentValue && field.currentValue.isChecked;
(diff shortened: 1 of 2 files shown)


src/modules/Preservation/views/ScopeOverview/TagFlyout/PreservationTab/Requirements.tsx
```diff
@@ -166,26 +165,43 @@ const Requirements = ({
         return isReadyToBePreserved;
     };
 
+    const checkValue = (requirementId: number, field: TagRequirementField): boolean | undefined => {
```
kfaln, line 168: Could this be simplified considerably? Something like this?

```
const isCheckBoxChecked = (field: TagRequirementField): boolean => field.currentValue && field.currentValue.isChecked;
```

This is basically the same check as the code removed from `RequirementCheckBoxField`
Why find the Field when it's already part of the function arguments...

bratli: This existing code did not work, no.. :(

bratli: The reason: this only works if the checkbox is not checked to begin with: thus, field.current.value is undefined. This means that it will always be undefined, and so I think checkbox has some logic that catches if it is clicked or not.
If it is checked, field.currentValue.isChecked never changes, hence, the checkbox UI never changes.

bratli: I am not finding field, but checking if the field is in requirement.checkboxValues, and using this if it exists. If it is not in requirement.checkbox values, the code that was used before is used, `field.currentValue && field.currentValue.isChecked`.

kfaln: The bug is probably due to a change made in the generic Checkbox component, perhaps related to using it with filters. This fix for requirements should be ok.

PR #114, opened 2020-05-28
Feature/73943 ui journey library add update void unvoid delete

arildbjerkenes:
We still miss endpoints for delete and void of steps. Will be added later.

### DIFF
--- a/src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
+++ b/src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
@@ -113,11 +113,10 @@ const PreservationJourney = (props: PreservationJourneyProps): JSX.Element => {
     }, []);
 
 
-    //todo: bare gjøre direkte i useEffekt? 
-    const getJourney = async (): Promise<void> => {
+    const getJourney = async (journeyId: number): Promise<void> => {
         setIsLoading(true);
         try {
-            await preservationApiClient.getJourney(props.journeyId).then(
+            await preservationApiClient.getJourney(journeyId).then(
                 (response) => {
                     setJourney(response);
                     setNewJourney(cloneJourney(response));
@@ -132,7 +131,7 @@ const PreservationJourney = (props: PreservationJourneyProps): JSX.Element => {
 
     useEffect((): void => {
         if (props.journeyId) {
-            getJourney();
+            getJourney(props.journeyId);
             setIsEditMode(true);
         } else {
             setJourney(null);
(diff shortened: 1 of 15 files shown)


src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
```diff
@@ -0,0 +1,501 @@
...
+            },
+            rowVersion: ''
+        };
+
+        newJourney.steps.push(newStep);
+
+        setNewJourney(cloneJourney(newJourney));
+    };
+
+    const swapSteps = async (stepA: Step, stepB: Step): Promise<void> => {
```
bratli, line 331: Maybe swap steps can wait to be actually saved until the user saves? This implemetation causes a reload of the component every time the user changes te order, causeing the loading symbol to pop up every time

bratli: Doing this will also allow the user to change order of newly added steps, without saving first. Now, if the user adds a step and tries to move it up before saving, they get an error that doesnt tell the user much

arildbjerkenes: I haven't been able to test up/down yet myself, because the endpoints had issues. I will test and fix. We agreed on just saving immediately, to simplify implementation. I think if the endpoint return 'sortKey', it would be easier for us to implement regular save.

src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
```diff
@@ -0,0 +1,501 @@
...
+                    const responsibleSelectItem: SelectItem | undefined = mappedResponsibles.find(resp => resp.value === step.responsible.code);
+
+
+                    return (
+                        <React.Fragment key={`step._${index}`}>
+
+                            <FormFieldSpacer>
+
+                                <SelectInput
+                                    onChange={(value): void => setModeValue(value, index)}
```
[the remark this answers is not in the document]
arildbjerkenes: Fixed

src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
```diff
@@ -0,0 +1,501 @@
...
+    const moveStepDown = (stepIndex: number): void => {
+        const steps = newJourney.steps;
+        if (steps.length > 1 && stepIndex + 1 < steps.length) {
+            swapSteps(steps[stepIndex], steps[stepIndex + 1]);
+        }
+    };
+
+    const moveStepUp = (stepIndex: number): void => {
+        const steps = newJourney.steps;
+        if (steps.length > 1 && stepIndex - 1 > 0) {
```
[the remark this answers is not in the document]
arildbjerkenes: Nice!

src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
arildbjerkenes: I have fixed the up/down button so that they are disabled for new steps. We can create a PBI for the future, to implement a new endpoint that just receives the whole journey, so that all changes can be taken care of by backend.

PR #118, opened 2020-06-02
Bug/missing classification on disciplines endpoint

bratli:
Classifications array on disciplines endpoint


src/modules/PlantConfig/http/LibraryApiClient.ts
```diff
@@ -166,9 +166,17 @@ class LibraryApiClient extends ApiClient {
      *
      * @param setRequestCanceller Returns a function that can be called to cancel the request
      */
-    async getDisciplines(setRequestCanceller?: RequestCanceler): Promise<DisciplineResponse[]> {
+    async getDisciplines(classifications: string[], setRequestCanceller?: RequestCanceler): Promise<DisciplineResponse[]> {
         const endpoint = '/Disciplines';
-        const settings: AxiosRequestConfig = {};
+        const qs = require('qs');
```
Cnordbo, line 171: move this to the top of the file. and use`import Qs from 'qs'`

src/modules/PlantConfig/http/LibraryApiClient.ts
```diff
@@ -3,6 +3,7 @@ import { AxiosRequestConfig } from 'axios';
 import { IAuthService } from '../../../auth/AuthService';
 import { RequestCanceler } from '../../../http/HttpClient';
 import {ErrorResponse, RegisterResponse, TagFunctionResponse} from './LibraryApiClient.types';
+import Qs from 'Qs';
```
Cnordbo, line 6: should always match the casing of the package in the 'path'.

```node
import Qs from 'qs'

PR #128, opened 2020-06-04
Feature/73457   filter improvements

bratli:
[73457](https://statoildeveloper.visualstudio.com/Statoil.NextGenProCoSys/_boards/board/t/Statoil.NextGenProCoSys%20Team/Backlog%20items/?workitem=73457)

- Checkboxes/radio groups are indented in a bit
- Tooltip on active filter
- Filter icon has green background when active
- Icons next to filter groups
- Title on filter groups has background if active
- Reset filter only active if filter is applied
- Preserve this week button on scope overview is closer to 'Start', as in Figma


src/modules/Preservation/views/ScopeOverview/ScopeFilter/ScopeFilter.tsx
```diff
@@ -373,17 +390,17 @@ const ScopeFilter = ({
...
+            <CheckboxFilter title='Requirements' filterValues={requirements} tagListFilterParam='requirementTypeIds' onCheckboxFilterChange={onCheckboxFilterChange} itemsChecked={tagListFilter.requirementTypeIds} icon={'pressure'}/>
+            <CheckboxFilter title='Tag Functions' filterValues={tagFunctions} tagListFilterParam='tagFunctionCodes' onCheckboxFilterChange={onCheckboxFilterChange} itemsChecked={tagListFilter.tagFunctionCodes} icon={'verticle_split'}/>
+            <CheckboxFilter title='Discipline' filterValues={disciplines} tagListFilterParam='disciplineCodes' onCheckboxFilterChange={onCheckboxFilterChange} itemsChecked={tagListFilter.disciplineCodes} icon={'category'} />
+            <MultiSelectFilter headerLabel="Responsible" items={responsibles} onChange={responsibleFilterUpdated} inputLabel="Responsible" inputPlaceholder="Select responsible" icon={<EdsIcon name='person' />} />
+            <MultiSelectFilter headerLabel="Area (on-site)" items={areas} onChange={areaFilterUpdated} inputLabel="Area" inputPlaceholder="Select area" icon={<AreaIcon />}  />
```
Cnordbo, line 403: Can these follow the same pattern as CheckboxFilter? (specify just the name, not the component?)

```js
<MultiSelectFilter icon='person' />....
```

bratli: The reason why I had to do this differently is because EDS does not have the icon for area... Any fixes that come to mind?

Cnordbo: We should probably ask the EDS designers to be careful when using icons outside EDS or remember, as a reminder, and/or to incorporate them in EDS so that they are available to all systems. 

But for now, i dont have a better idea.

PR #129, opened 2020-06-04
Remove toolbar text. Not necessary, all selected tags show to right.

bratli:
Reason for fix:
Number of selected tags displayed on top of table showed only number of selected tags on that search. Could be confusing for user. Lotte says just remove toolbar text, as number of selected tags is show on right,


Cnordbo:
Remove the test that is associated with the functionality, and all is good =)

PR #131, opened 2020-06-05
Feature/72397   library framework should support url that will bring up the page

bratli:
Update URL based on library item clicked.

Note:
-if you have a url with parameters filled out and reload or copy/pase it on a new page, the parameters are cleared because the tree view will not open to that.
-url only updated when clicking on an item in the treeview, not just by expanding a part.


Cnordbo:
The issue you are facing here (URLs not opening) is why this is setup as a larger task, this must be addressed before merging.

PR #133, opened 2020-06-08
Feature/74323   po tag

bratli:
Create #PO tags by choosing Supplier in the 'Area' dropdown.
Instead of Area type the user must choose PO/CO. This is not optional.

Supplier is chosen as step on page two of wizard when Journey is selected, as it says in PBI that it does not make sense to choose any other step. Hence, if the user selects a Journey that does not have Supplier as a step, they cannot use this Journey, and must choose a Journey that has Supplier as a step.


bratli:
Merk description her og at jeg har forstått logikk i Wizard side 2 riktig.

PR #144, opened 2020-06-18
Feature/74384 ui add support for migration

arildbjerkenes:


src/modules/Preservation/views/AddScope/AddScope.style.ts
```diff
@@ -3,14 +3,15 @@ import { tokens } from '@equinor/eds-tokens';
 
 export const Container = styled.div`
     display: flex;
+    overflow-x: scroll;
```
bratli, line 6: Tror den nye stylingen din har fororsaket litt problemer? Se Create Area Tag, der er det en stor disabled scrollbar. (Scrollbar på alle AddScopene, men noen enabled og noen disabled).
Migrate siden ser rar ut. Tabell tar opp for mye plass, må scrolle for å se selected tags, og selected-tags-komponenten er veldig smal.

src/modules/Preservation/views/AddScope/AddScope.tsx
```diff
@@ -304,7 +380,26 @@ const AddScope = (): JSX.Element => {
...
+                        selectedTags={selectedTags}
+                        migrationTableData={migrationTableData}
+                        isLoading={isLoading}
+                        addScopeMethod={addScopeMethod}
+                        removeTag={removeSelectedTagMigration}
+                        removeFromMigrationScope={removeFromMigrationScope}
+                    />
+                    <Divider />
+                    <SelectedTags>
+                        <TagDetails selectedTags={selectedTags} showMCPkg={false} collapsed={false} />
```
bratli, line 398: send inn removeTag prop for å få delete

bratli: ikke send inn collapse. ser rart ut når man velger flere tags

src/modules/Preservation/views/AddScope/AddScope.style.ts
arildbjerkenes: Fixed.

src/modules/Preservation/views/AddScope/AddScope.tsx
arildbjerkenes: Jepp, forglemmelse.

src/modules/Preservation/views/AddScope/SelectMigrateTags/SelectMigrateTags.tsx
```diff
@@ -0,0 +1,174 @@
...
+                <div>{project.description}</div>
+            </Header>
+            <TopContainer>
+                <InnerContainer>
+                    <TagsHeader>Select the tags that should be migrated, and click &apos;next&apos;</TagsHeader>
+                </InnerContainer>
+                <ButtonsContainer>
+                    <Button onClick={cancel} variant='outlined' >Cancel</Button>
+                    <ButtonSeparator />
+                    <Button onClick={props.removeFromMigrationScope} disabled={props.selectedTags.length === 0}>Remove from migration scope</Button>
```
bratli, line 129: Is this button neccessary? We dont have this in autoscope and manual... Should be consistent.
And should rename to 'Clear migration scope' if we need it?
If the user has selected tags, and filters so that we see only one tag, selects it, and click 'Remove from migration scope', this button is on the table component, giving the user the impression that maybe only the tag shown in the filtered table will be removed. Therefore Clear, since this is what we are doing.

bratli, line 129: Is this button neccessary? We dont have this in autoscope and manual... Should be consistent.
And should rename to 'Clear migration scope' if we need it?
If the user has selected tags, and filters so that we see only one tag, selects it, and click 'Remove from migration scope', this button is on the table component, giving the user the impression that maybe only the tag shown in the filtered table will be removed. Therefore Clear, since this is what we are doing.

bratli:
Should this handle the fact that #PO-tags should only be on 'Supplier' step any way?

src/modules/Preservation/views/AddScope/SelectMigrateTags/SelectMigrateTags.tsx
arildbjerkenes: The remove button actually sets tags in old preservation to 'migrated'. This prevents the tags from appearing again in this list. The tags have not really been migrated. Then have been removed from the preservation scope.  I think 'clear' will be strange here. 

We could consider å dialogbox, to list all tags that will be removed.

PR #145, opened 2020-06-21
Added administration of modes

arildbjerkenes:
The mode endpoints misses 'isVoided', so voided modes are currently not shown as voided.

### DIFF
--- a/src/modules/PlantConfig/views/Library/Mode/Mode.tsx
+++ b/src/modules/PlantConfig/views/Library/Mode/Mode.tsx
@@ -136,6 +136,7 @@ const Mode = (props: ModeProps): JSX.Element => {
             try {
                 await preservationApiClient.voidMode(mode.id, mode.rowVersion);
                 getMode(mode.id);
+                props.setDirtyLibraryType();
                 showSnackbarNotification('Mode is voided.', 5000);
             } catch (error) {
                 console.error('Error occured when trying to void mode: ', error.messsage, error.data);
@@ -151,6 +152,7 @@ const Mode = (props: ModeProps): JSX.Element => {
             try {
                 await preservationApiClient.unvoidMode(mode.id, mode.rowVersion);
                 getMode(mode.id);
+                props.setDirtyLibraryType();
                 showSnackbarNotification('Mode is unvoided.', 5000);
             } catch (error) {
                 console.error('Error occured when trying to unvoid mode: ', error.messsage, error.data);
@@ -219,7 +221,7 @@ const Mode = (props: ModeProps): JSX.Element => {
                 </Button>
             </ButtonContainer>
 
-            <InputContainer>
+            <InputContainer style={{ width: '280px' }}>
                 <TextField
                     id={'title'}
                     label='Title for this mode'
@@ -233,6 +235,7 @@ const Mode = (props: ModeProps): JSX.Element => {
             <InputContainer>
                 <Checkbox
                     checked={newMode.forSupplier}
+                    disabled={newMode.isVoided}
                     onChange={(checked: boolean): void => {
                         setForSupplierValue(checked);
                     }}
--- a/src/modules/PlantConfig/views/Library/Mode/Mode.tsx
+++ b/src/modules/PlantConfig/views/Library/Mode/Mode.tsx
@@ -19,6 +19,7 @@ interface ModeItem {
 
 type ModeProps = {
     modeId: number;
+    setDirtyLibraryType: () => void;
 };
 
 const Mode = (props: ModeProps): JSX.Element => {
@@ -72,6 +73,7 @@ const Mode = (props: ModeProps): JSX.Element => {
         try {
             const modeId = await preservationApiClient.addMode(newMode.title, newMode.forSupplier);
             getMode(modeId);
+            props.setDirtyLibraryType();
             showSnackbarNotification('New mode is saved.', 5000);
         } catch (error) {
             console.error('Add mode failed: ', error.messsage, error.data);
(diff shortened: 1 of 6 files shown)


src/modules/PlantConfig/views/Library/Mode/Mode.tsx
```diff
@@ -0,0 +1,247 @@
...
+                    Cancel
+                </Button>
+                <ButtonSpacer />
+                <Button onClick={handleSave} disabled={newMode.isVoided || !isDirty}>
+                    Save
+                </Button>
+            </ButtonContainer>
+
+            <InputContainer>
+                <TextField
```
bratli, line 220: Make the text field smaller. It goes accross the whole page now.

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
```diff
@@ -0,0 +1,247 @@
...
+                    label='Title for this mode'
+                    value={newMode.title}
+                    onChange={(e: React.ChangeEvent<HTMLInputElement>): void => { setTitleValue(e.target.value); }}
+                    placeholder="Write Here"
+                    disabled={newMode.isVoided}
+                />
+            </InputContainer>
+
+            <InputContainer>
+                <Checkbox
```
bratli, line 231: Tooltip on checkbox, see Lotte's design

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
```diff
@@ -0,0 +1,247 @@
...
+                    </Button>
+                }
+
+                {!newMode.isVoided &&
+                    <Button variant="outlined" onClick={voidMode}>
+                        Void
+                    </Button>
+                }
+                <ButtonSpacer />
+                <Button variant="outlined" onClick={cancel} disabled={newMode.isVoided}>
```
bratli, line 210: Cancel is not part of design. Should it be here?

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
```diff
@@ -0,0 +1,247 @@
...
+    const cancel = (): void => {
+        if (isDirty && !confirm('Do you want to cancel changes without saving?')) {
+            return;
+        }
+        setMode(null);
+        setIsEditMode(false);
+    };
+
+    const voidMode = async (): Promise<void> => {
+        if (mode) {
```
bratli, line 131: Voided mode should have some indication that it is coided. Should not be able to check/uncheck textbox or edit text field. Should have button that says Unvoid instead of void

bratli:
Also, since mode has just a textbox and checkbox, I think the Save button can toggle between enabled/disabled based on if there actually are any chanegs to save. Eg, mode har checkbox unchecked originally. The user checks the checkbox -> save enabled. The user unchecks the textbox again -> save disabled

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
arildbjerkenes: fixed

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
arildbjerkenes: Design is not updated. There is a Cancel button in other similar screens.

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
arildbjerkenes: My intention was to fix the impl of void when this is supported by the endpoint. But I can fix this now, anyway.

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
```diff
@@ -0,0 +1,250 @@
...
+            setIsLoading(true);
+            try {
+                await preservationApiClient.voidMode(mode.id, mode.rowVersion);
```
kfaln, line 137: Should probably throw in `props.setDirtyLibraryType();` here (since voided node has a different rendering in the tree).

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
```diff
@@ -0,0 +1,250 @@
...
+            setIsLoading(false);
+        }
+    };
+
+    const unvoidMode = async (): Promise<void> => {
+        if (mode) {
+            setIsLoading(true);
+            try {
+                await preservationApiClient.unvoidMode(mode.id, mode.rowVersion);
+                getMode(mode.id);
```
kfaln, line 153: Should probably throw in `props.setDirtyLibraryType();` here (since voided node has a different rendering in the tree).

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
arildbjerkenes: Yes

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
arildbjerkenes: Yes

src/modules/PlantConfig/views/Library/Mode/Mode.tsx
arildbjerkenes: Checkbox component does not support this. We should not spend time on the is now. This is not important. Users will modify the modes very rarely, and will probably never change that SUPPLIER mode has this flag set.

PR #149, opened 2020-06-23
Feature/71765   ui singlesignout via microsoft

bratli:
Singlesignout. Clear session and local storage.


Cnordbo:
Move implementation to `logout` in `AuthService`

PR #156, opened 2020-06-25
74965 - Interpret uniform messages on 400

kfaln:
Parse input and business validation error messages.

### DIFF
--- a/src/modules/Preservation/http/PreservationApiClient.ts
+++ b/src/modules/Preservation/http/PreservationApiClient.ts
@@ -429,16 +429,20 @@ function getPreservationApiError(error: AxiosError): PreservationApiError {
         return new PreservationApiError(error.response.data, error.response);
     }
     if (error.response.status == 400) {
-        // input and business validation errors
-        let validationErrorMessage = error.response.data.title;
-        const validationErrors = error.response.data.errors;
+        try {
+            // input and business validation errors
+            let validationErrorMessage = error.response.data.title;
+            const validationErrors = error.response.data.errors;
 
-        for (const validatedField in validationErrors) {
-            const fieldErrors = validationErrors[validatedField].join(' | ');
-            validationErrorMessage += ` ${validatedField}: ${fieldErrors} `;
+            for (const validatedField in validationErrors) {
+                const fieldErrors = validationErrors[validatedField].join(' | ');
+                validationErrorMessage += ` ${validatedField}: ${fieldErrors} `;
+            }
+
+            return new PreservationApiError(validationErrorMessage, error.response);
+        } catch (exception) {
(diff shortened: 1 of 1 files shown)


src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -428,6 +428,22 @@ function getPreservationApiError(error: AxiosError): PreservationApiError {
     if (error.response.status == 404) {
         return new PreservationApiError(error.response.data, error.response);
     }
+    if (error.response.status == 400) {
+        try {
+            // input and business validation errors
+            let validationErrorMessage = error.response.data.title;
```
Cnordbo, line 434: No sure if this gives us any relevant context ("validation error occured" or " Bussiness validation errors: X " )

src/modules/Preservation/http/PreservationApiClient.ts
```diff
@@ -428,6 +428,22 @@ function getPreservationApiError(error: AxiosError): PreservationApiError {
...
+            const validationErrors = error.response.data.errors;
+
+            for (const validatedField in validationErrors) {
+                const fieldErrors = validationErrors[validatedField].join(' | ');
+                validationErrorMessage += ` ${validatedField}: ${fieldErrors} `;
```
Cnordbo, line 439: the field doesnt necessarily give the user any context, currently we should probably rely on descriptive errormessage as fieldnames are internal to backend.

src/modules/Preservation/http/PreservationApiClient.ts
kfaln: I guess the back-end should provide more context-specific messages, if the need arises. Front-end will display whatever it comes up with.

src/modules/Preservation/http/PreservationApiClient.ts
kfaln: Perhaps, but I think you will get a pretty close match on the front VS back field name most of the times. Here is an example:

Remark: The length of 'Remark' must be 255 characters or fewer. You entered 339 characters.

kfaln: The field is also a part of the error message, so again it's up to back-end to supply niceness to the user.

Cnordbo: The field in the error message is handled by the implementor as a string error message. 
The field ( as a key ) is just the property of an object. 

ex. "StepId: Step does not exist" 
ex. "RequirementDescription: Description can not be more then 255 characters"

kfaln: Removed the internal field name from error message.

PR #159, opened 2020-06-29
74993: Link to tag in main from tag flyout

bratli:
Link to tag in main from tag flyout


src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -136,9 +157,11 @@ const TagFlyout = ({
                 </HeaderNotification>
             }
             <Header>
-                <h1>
-                    {tagDetails ? tagDetails.tagNo : '-'}
-                </h1>
+                <TagNoContainer onClick={goToTag}>
```
kfaln, line 160: How snappy is this.. does it seem unresponsive when the use clicks the link? Would it be better if the tagId was preloaded when the tag flyout is opened? If you feel that the performance is ok then it's fine.

bratli: I agree, the performance is better when getting tagId on load. Fixed.

PR #166, opened 2020-06-30
Fixed routing error. Only link on standard tags, not area tags

bratli:


src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -75,7 +76,7 @@ const TagFlyout = ({
         return (): void => {
             requestCancellor && requestCancellor();
         };
-    }, [tagDetails]);
+    }, [isStandardTag]);
```
kfaln, line 79: Do we need the separate state variable? I think I would prefer to use tagDetails and check on tagDetails.tagType where needed... perhaps add a helper function something like

`const isStandardTag = (): boolean => tagDetails && tagDetails.tagType === 'Standard';`

src/modules/Preservation/views/ScopeOverview/TagFlyout/TagFlyout.tsx
```diff
@@ -49,18 +48,24 @@ const TagFlyout = ({
...
             setTagDetails(details);
-            setIsStandardTag(details.tagType == 'Standard');
         }
         catch (error) {
             console.error(`Get TagDetails failed: ${error.message}`);
             showSnackbarNotification(error.message, 5000, true);
         }
     };
 
+    const isStandardTag = (): boolean => {
```
kfaln, line 58: Great :) Just a tip - functions like these can be shortened like this:

`const isStandardTag = (): boolean => tagDetails ? tagDetails.tagType === 'Standard' : false;`

PR #167, opened 2020-06-30
Added link to old preservation

arildbjerkenes:
I need input from you, Christer, if this is the way to do this.  'OldPreseration' or 'oldPreservation'?

### DIFF
--- a/src/modules/Preservation/index.tsx
+++ b/src/modules/Preservation/index.tsx
@@ -2,18 +2,26 @@ import { Route, BrowserRouter as Router, Switch, useRouteMatch } from 'react-rou
 
 import AddScope from './views/AddScope/AddScope';
 import EditTagProperties from './views/EditTagProperties/EditTagProperties';
-import { Container } from './style';
+import { Container, OldPreservationLink } from './style';
 import { PreservationContextProvider } from './context/PreservationContext';
 import React from 'react';
 import ScopeOverview from './views/ScopeOverview/ScopeOverview';
 import withAccessControl from '../../core/security/withAccessControl';
+import { Typography } from '@equinor/eds-core-react';
 
 const Preservation = (): JSX.Element => {
+
+    const switchToOld = (): void => {
+        window.location.href = 'OldPreservation';
+    };
 
     const { url } = useRouteMatch();
     return (
         <PreservationContextProvider>
             <Container>
+                <OldPreservationLink title='To work on preservation scope not yet migrated.' onClick={switchToOld}>
+                    <Typography bold variant="caption" >Switch to old system</Typography>
+                </OldPreservationLink>
                 <Router basename={url}>
                     <Switch>
                         <Route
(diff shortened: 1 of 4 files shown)


arildbjerkenes:
I will add project to the link...

src/modules/Preservation/index.tsx
```diff
@@ -2,18 +2,26 @@ import { Route, BrowserRouter as Router, Switch, useRouteMatch } from 'react-rou
...
 
+    const switchToOld = (): void => {
+        window.location.href = 'OldPreservation';
+    };
+
     const { url } = useRouteMatch();
     return (
         <PreservationContextProvider>
             <Container>
+                <OldPreservationLink title='To work on preservation scope not yet migrated.' onClick={switchToOld}>
```
kfaln, line 22: The link is visible in all views, is that wanted or just on scope overview? 
It also "floats" over the tag filter (may be ok, I don't know).

src/modules/Preservation/style.ts
```diff
@@ -1,7 +1,8 @@
 import styled from 'styled-components';
+import { tokens } from '@equinor/eds-tokens';
 
 export const Container = styled.div`
-    margin: var(--margin-module--top) var(--margin-module--right);
```
kfaln, line 4: This leads to some unwanted spacing in the "separators" used for instance in filter and "selected tags".

image

image

src/modules/Preservation/index.tsx
Cnordbo, line 22: a standard `a href` element can be used, which would remove the need for an extra callback function to trigger the redirect. 

(switch out `div` with `a` in the styled component)

arildbjerkenes: Agree, lets have it only for scope list.

arildbjerkenes: yes!

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -454,10 +455,17 @@ const ScopeOverview: React.FC = (): JSX.Element => {
     return (
         <Container>
             <ContentContainer>
+                <OldPreservationLink>
+                    <Typography bold variant="caption" >
+                        <Tooltip title='To work on preservation scope not yet migrated.' enterDelay={200} enterNextDelay={100} arrow={true}>
+                            <a href="OldPreservation">Switch to old system</a>
```
kfaln, line 461: Url goes to "....../OSEBERG_C/Preservation/OldPreservation". Is this correct, or should it be "....../OSEBERG_C/OldPreservation". If so, it can be fixed with `href="../OldPreservation"`

arildbjerkenes: ?? This is not how it looks locally, for me.

kfaln: ok now, probably local caching issue

PR #177, opened 2020-07-01
74972: Voided/Unvoided button updating

bratli:
Refactored requirement selector. Tag function updated when voided/unvoided. Cannot edit requirements on voided tag funciton.

### DIFF
--- a/src/modules/Preservation/components/RequirementsSelector/RequirementsSelector.tsx
+++ b/src/modules/Preservation/components/RequirementsSelector/RequirementsSelector.tsx
@@ -26,7 +26,7 @@ type RequirementsSelectorProps = {
     requirementTypes: RequirementType[];
     onChange?: (listOfRequirements: OnChangeRequirementData[]) => Promise<void> | void;
     requirements: RequirementFormInput[];
-    disableActions?: boolean;
+    disabled?: boolean;
 }
 
 interface RequirementFormInput {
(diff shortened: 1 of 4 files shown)


src/modules/Preservation/components/RequirementsSelector/RequirementsSelector.tsx
```diff
@@ -43,9 +42,13 @@ interface RequirementFormInput {
 
 const validWeekIntervals = [1, 2, 4, 6, 8, 12, 16, 24, 52];
 
-const RequirementsSelector = (props: RequirementsSelectorProps): JSX.Element => {
+const RequirementsSelector = ({
+    requirementTypes,
+    onChange,
+    requirements,
+    disableActions = false }: RequirementsSelectorProps): JSX.Element => {
```
Cnordbo, line 50: The use of `props` was done on purpose to avoid "local vs external" naming and be clear about what comes in as external properties vs what is manipulated internally.

src/modules/Preservation/components/RequirementsSelector/RequirementsSelector.tsx
```diff
@@ -43,9 +42,13 @@ interface RequirementFormInput {
...
```
Cnordbo, line 49: i think `disabled` should be the standard used trough all components, when disabled. (just like standard HTML elements)

PR #183, opened 2020-07-02
Bug/long select component max height scrollable

bratli:
Optional prop on selected to make list scrollable if too long

### DIFF
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -20,7 +20,7 @@ type SelectProps = {
     children: ReactNode;
     label?: string;
     isVoided?: boolean;
-    hasMaxHeight?: boolean;
+    maxHeight?: string;
 };
 
 const KEYCODE_ENTER = 13;
@@ -35,7 +35,7 @@ const Select = ({
     children,
     label,
     isVoided = false,
-    hasMaxHeight
+    maxHeight
 }: SelectProps): JSX.Element => {
     const [isOpen, setIsOpen] = useState(false);
     const containerRef = useRef<HTMLDivElement>(null);
@@ -133,7 +133,7 @@ const Select = ({
     };
 
     return (
-        <Container ref={containerRef} hasMaxHeight={hasMaxHeight}>
+        <Container ref={containerRef} maxHeight={maxHeight}>
             <Label isVoided={isVoided}>{label}</Label>
             <DropdownButton
                 isVoided={isVoided}
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -20,7 +20,7 @@ type SelectProps = {
     children: ReactNode;
     label?: string;
     isVoided?: boolean;
-    maxHeight?: boolean;
+    hasMaxHeight?: boolean;
 };
 
 const KEYCODE_ENTER = 13;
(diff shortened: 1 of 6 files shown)


src/components/Select/index.tsx
```diff
@@ -33,7 +34,8 @@ const Select = ({
     },
     children,
     label,
-    isVoided = false
+    isVoided = false,
+    maxHeight
```
kfaln, line 38: By the name it sounds like you should provide a maxHeight to the component. Perhaps rename it to "hasMaxHeight" or something?

src/components/Select/style.ts
```diff
@@ -2,27 +2,37 @@ import styled, { css } from 'styled-components';
...
-        li:only-child > div {
-            box-shadow: 0px 1px 5px rgba(0,0,0,0.2), 0px 3px 4px rgba(0,0,0,0.12), 0px 2px 4px rgba(0,0,0,0.14);
-        }
+        ${(props): any => props.maxHeight && css`
+            max-height: 300px;
+            overflow-y: auto;
+            box-shadow: ${tokens.elevation.raised};
+        `}
 
+        ${(props): any => !props.maxHeight && css`
```
kfaln, line 24: Not sure I understand why you need to apply this styling only when maxHeight is false... it seems to work ok without the maxHeight check. Also, when maxHeight is true, you lose a horizontal separator between the elements in a list.. almost invisible, but it's there :)

src/components/Select/style.ts
```diff
@@ -2,27 +2,37 @@ import styled, { css } from 'styled-components';
...
+    maxHeight?: boolean;
+}
 
-export const Container = styled.div`
+export const Container = styled.div<ContainerProps>`
     position: relative;
     ul {
         position: absolute;
         background-color: transparent;
         border-radius: 4px;
```
kfaln, line 14: Border-radius does not seem to have an effect for maxHeight = false lists. Not sure if this is something new, just noticed that they don't have rounded corners.

Cnordbo:
Instead of using `hasMaxHeight` you should just use `maxHeight` and use it as `maxHeight="300px" (or just 300), and apply the extra styling based on that. 

Can also be implemented as a default prop so that no dropdown extend the 300px.

bratli:
> Instead of using hasMaxHeight you can just use maxHeight and use it as `maxHeight="300px" (or just 300), and apply the extra styling based on that.
> 
> Can also be implemented as a default prop so that no dropdown extend the 300px.

@Cnordbo 
I don't see how that is better? Do you mean so that the user has the option of setting a max hight different than 300px? eg 500px for a select?

The reason why I have not made select have max height by default is because of the requirement selector has some issues with it. If they want this scrollable this can be done, but I think it deserves its own card/bug.

src/components/Select/style.ts
```diff
@@ -2,27 +2,42 @@ import styled, { css } from 'styled-components';
...
         z-index: 100;
         white-space: nowrap;
+
+        ${(props): any => props.hasMaxHeight && css`
+            max-height: 300px;
+            overflow-y: auto;
+            box-shadow: ${tokens.elevation.raised};
```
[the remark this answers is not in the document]
bratli: Set shadow on ul-element. If not the dropdown gets no shadow at all.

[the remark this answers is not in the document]
Cnordbo: Sure, i just dont get why it would only have the shadow if maxheight is set. 
Seems like the iffy can be removed, and just apply shadow on the UL always? or is it specified to be something else?

PR #188, opened 2020-07-06
75228: Fix/table fix

bratli:


src/modules/Preservation/views/ScopeOverview/ScopeOverview.style.ts
```diff
@@ -11,8 +11,8 @@ export const ContentContainer = styled.div`
     flex-direction: column;
     overflow-x: hidden;
     width: 100%;
-    overflow: visible;
     margin-top: -16px;
+    min-height: 400px; /* min-height to ensure that project dropdown (max 300px) is not cut off if empty table */
```
Cnordbo, line 15: For information: 
This issue can probably be fixed by elevating the z-index of the project selector, and would in many cases be a more suitable fix as that would better support responsive designs.

bratli:
> For information:
> This issue can probably be fixed by elevating the z-index of the project selector, and would in many cases be a more suitable fix as that would better support responsive designs.

Elevating z-index did not fix this issue, as it does not "overwrite" the overflow:hidden property of the parent element.... I did not find an elegant fix here, because I want to have overflow-x: hidden and overflow-y: visible to allow the filter to get fit into the screen, but the dropdown to overflow the parent div. However, combining these two property makes the overflow-y: visible act as overflow: auto.

PR #192, opened 2020-07-07
Added global dirty for components

arildbjerkenes:
I have done som 'proof of concept' here. I only handle switching between library items. We need to support switching to other tabs and outside the application. 

Instead of just providing a set object ('dirtyComponents'), we could offer some methods to add/remove dirty flag on components.


Cnordbo:
I would suggest abbandoning this for e more generic API that is more inline with React and state updates in #230

PR #196, opened 2020-07-09
75227 - UI - Action bell icon is missing from tag list

kfaln:
Added action column to scope list.
Setting scope list dirty on action create and close.

### DIFF
--- a/src/modules/Preservation/http/PreservationApiClient.ts
+++ b/src/modules/Preservation/http/PreservationApiClient.ts
@@ -10,6 +10,7 @@ const scopes = JSON.parse(Settings.externalResources.preservationApi.scope.repla
 interface PreservedTagResponse {
     maxAvailable: number;
     tags: [{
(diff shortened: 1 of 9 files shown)


bratli:
Don't know who created the actions tab component, but is it in the design to have a square ghost button on the arrow to expand an action? I would expect it to be a circular green background on hover, like other places (eg. Selected tags component when adding to scope).
If it is an easy fix, include it in this PR? Not a big deal I guess since it is not specified/reported.

bratli:
> Don't know who created the actions tab component, but is it in the design to have a square ghost button on the arrow to expand an action? I would expect it to be a circular green background on hover, like other places (eg. Selected tags component when adding to scope).
> If it is an easy fix, include it in this PR? Not a big deal I guess since it is not specified/reported.

I see some other buggy things in Action tab, so I will just create a new bug.

PR #208, opened 2020-07-24
Feature/74836   standardize buttons scope overview buttons

bratli:
Buttons on header level on scope overview are only enabled if any tags that meet the requirement of that button's action are selected.
Eg. if we select a tag and it is not ready to be preserved, then the Ready to be preserved button is not enabled.

Added section on void/unvoid dialog for tags that are not able to be voided/unvoided

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
@@ -136,10 +136,10 @@ const ScopeOverview: React.FC = (): JSX.Element => {
     useEffect(() => {
         setVoidedTagsSelected(selectedTags.find(t => t.isVoided) ? true : false);
         setUnvoidedTagsSelected(selectedTags.find(t => !t.isVoided) ? true : false);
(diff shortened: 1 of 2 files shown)


src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -374,22 +383,33 @@ const ScopeOverview: React.FC = (): JSX.Element => {
...
+        //     const newTag: PreservedTag = { ...tag };
+        //     if (!tag.isVoided && voiding) {
+        //         voidableTags.push(newTag);
+        //     } else if (tag.isVoided && !voiding) {
+        //         unvoidableTags.push(newTag);
+        //     }
+        // });
+        // const voidButton = voidableTags.length > 0 ? 'Void' : 'Unvoid';
+        // const voidFunc = voidableTags.length > 0 ? voidTags : unVoidTags;
+        // const voidTitle = voidableTags.length > 0 ? 'Voiding following tags' : 'Unvoiding following tags';
```
arildbjerkenes, line 397: This should be removed.

PR #220, opened 2020-08-11
Add RFCC and RFOC fields on journey step

arildbjerkenes:


src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
```diff
@@ -671,6 +688,40 @@ const PreservationJourney = (props: PreservationJourneyProps): JSX.Element => {
...
+                                    <div style={{ display: 'flex', flexDirection: 'row' }}>
+                                        <FormFieldSpacer>
+                                            <Checkbox
+                                                checked={step.autoTransferRFCC}
+                                                disabled={newJourney.isVoided || step.isVoided || step.autoTransferRFOC}
+                                                onChange={(checked: boolean): void => {
+                                                    setStepAutoTransValue(checked, 'RFCC', index);
+                                                }}
+                                            >
+                                                <Typography title='asdf' variant='body_long'>RFCC</Typography>
```
paal123, line 706: asdf?

arildbjerkenes: fixed

src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
```diff
@@ -401,10 +411,19 @@ const PreservationJourney = (props: PreservationJourneyProps): JSX.Element => {
         }
     };
 
+    const setStepAutoTransValue = (value: string, index: number): void => {
+        if (newJourney.steps) {
+            setIsDirty(true);
+            newJourney.steps[index].autoTransferMethod = value;
+            setNewJourney(cloneJourney(newJourney));
```
paal123, line 418: A rather confusing line:
setNewJourney(cloneJourney(newJourney));

Could it be renamed to better say what is happening here?

arildbjerkenes: The reason we do this is that we cannot just change the value on a state variable. We have to call the setNewJourney method. We also have to create a clone of the object, because react will not accept this as a change unless it is a new object. We have used 'new<library type>' several places. I think we should keep this.

arildbjerkenes:
Fixed issue with disabling/enabling buttons.

PR #223, opened 2020-08-17
75247: Breadcrumbs in library

bratli:


arildbjerkenes (commented):
After discussions with Elisabeth, I think we should accept this implementation as a temporary solution. Later we should generalize the implementation so that breadcrumb always appears automatically on library detail pages, based on titles in treeview. The treeview component can have a function that returns the parents. Also, we can make the entries in the breadcrumb as links.

arildbjerkenes (changes requested):
There are no space between the border and the buttons on the right side of the detail window.

Cnordbo (approved):

PR #225, opened 2020-08-18
Bug/75381 ui bug filter search freetext

arildbjerkenes:


Cnordbo (approved):
Please be more considerate with just comitting "all changes" when its just empty lines. This causes files to have an irrelevant commit history

PR #229, opened 2020-08-24
Add possiblity to delete fields on req.def. that are not in use

arildbjerkenes:


bratli:
image
On this voided req def, delete on field is enabled. However, deleting does not enable save button. So maybe disable delete on field when entire req def is voided? Alternative is to be able to save voided req def when deleting field

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
```diff
@@ -609,7 +619,7 @@ const PreservationRequirementDefinition = (props: PreservationRequirementDefinit
                                         </Button>
                                     </>
                                 }
-                                {(field.id == null) &&
+                                {(field.id == null || (field.isVoided && !field.isInUse)) &&
                                     (<Button variant='ghost' title="Delete" onClick={(): void => deleteField(index)}>
```
bratli, line 623: If fields are removed using delete icon, and then press 'Save', fields are not really delete from req def. Is this solely because of backend?

bratli (changes requested):
See comment about enabled delete on voided req def

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
arildbjerkenes: Yes, backend is responsible for deleting fields that are removed.. I thought it would be in place by now...

arildbjerkenes:
> image
> On this voided req def, delete on field is enabled. However, deleting does not enable save button. So maybe disable delete on field when entire req def is voided? Alternative is to be able to save voided req def when deleting field

When I test this, the save button do get enabled when I delete a field.

bratli:
> > image
> > On this voided req def, delete on field is enabled. However, deleting does not enable save button. So maybe disable delete on field when entire req def is voided? Alternative is to be able to save voided req def when deleting field
> 
> When I test this, the save button do get enabled when I delete a field.

In the code, line 549,         <Button onClick={handleSave} disabled={newRequirementDefinition.isVoided || !isDirtyAndValid}>
So it says what I am saying, Save is always disabled when reqDef is voided.

bratli (changes requested):
See my response to your comment

arildbjerkenes:
This PR is dependent on Product Backlog Item 76377: Api - The isInUse flag on ReqDef and Delete of ReqDef.

arildbjerkenes:
> > > image
> > > On this voided req def, delete on field is enabled. However, deleting does not enable save button. So maybe disable delete on field when entire req def is voided? Alternative is to be able to save voided req def when deleting field
> > 
> > 
> > When I test this, the save button do get enabled when I delete a field.
> 
> In the code, line 459, <Button onClick={handleSave} **disabled={newRequirementDefinition.isVoided** || !isDirtyAndValid}>
> So it says what I am saying, Save is always disabled when reqDef is voided.

I misunderstood. Delete button on field should be voided here.

bratli (approved):

PR #231, opened 2020-08-26
77654: Update EDS package

bratli:
-Update EDS package. 
-Global color moved to body, need to add it where body does not reach. 
-Global size of icons inside buttons. Specifies new size where needed.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/ScopeOverview.style.ts
+++ b/src/modules/Preservation/views/ScopeOverview/ScopeOverview.style.ts
@@ -34,10 +34,6 @@ export const Header = styled.header`
         margin-right: calc(var(--grid-unit) * 2);
     }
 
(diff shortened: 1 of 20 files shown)


src/modules/Preservation/views/ScopeOverview/ScopeOverview.style.ts
```diff
@@ -125,3 +112,7 @@ export const OldPreservationLink = styled.div`
     text-decoration: underline;
     cursor: pointer;
 `;
+
+export const ButtonSpacer = styled.div`
+    margin-right: calc(var(--grid-unit) * 2);
+`;
```
Cnordbo, line 118: Avoid fluffing components as much as possible.

Cnordbo (approved):

PR #234, opened 2020-08-26
Feature/76105 ui delete of requirement definition2

arildbjerkenes:
This PR ended up including some additional one-liner fixes.

### DIFF
--- a/src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
+++ b/src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
@@ -392,6 +392,16 @@ const PreservationRequirementDefinition = (props: PreservationRequirementDefinit
         return true;
     };
 
+    const showDeleteFieldButton = (field: Field): boolean => {
+        if (requirementDefinition) {
+            const origField = requirementDefinition.fields.find((f) => f.id = field.id);
+            if (field.id == null || (origField && origField.isVoided && field.isVoided && !field.isInUse)) {
+                return true;
+            }
+        }
+        return false;
+    };
+
     if (isLoading) {
         return (<Container>
             <Breadcrumbs>{'Library / Preservation Requirements /'}</Breadcrumbs>
@@ -619,7 +629,7 @@ const PreservationRequirementDefinition = (props: PreservationRequirementDefinit
                                         </Button>
                                     </>
                                 }
-                                {(field.id == null || (field.isVoided && !field.isInUse)) &&
+                                {showDeleteFieldButton(field) &&
                                     (<Button disabled={newRequirementDefinition.isVoided} variant='ghost' title="Delete" onClick={(): void => deleteField(index)}>
                                         {deleteIcon}
                                     </Button>)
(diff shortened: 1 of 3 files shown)


src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
```diff
@@ -392,6 +391,16 @@ const PreservationRequirementDefinition = (props: PreservationRequirementDefinit
         return true;
     };
 
+    const showDeleteFieldButton = (field: Field): boolean => {
+        if (requirementDefinition) {
+            const origField = requirementDefinition.fields.find((f) => f.id = field.id);
```
bratli, line 396: I think origField is not a good variable name. Is it an existing field?
should be (f) => f.id === field.id, so with three ===

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
```diff
@@ -392,6 +391,16 @@ const PreservationRequirementDefinition = (props: PreservationRequirementDefinit
...
+            if (field.id == null || (origField && origField.isVoided && field.isVoided && !field.isInUse)) {
```
bratli, line 397: I think you only need to check if origField is voided, or if field is voided, not both. They are the same. Decide which one you want to use, and use the same to check for isInUse.

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
```diff
@@ -189,20 +189,19 @@ const PreservationRequirementDefinition = (props: PreservationRequirementDefinit
...
+
                             if (singleReqDef) {
                                 const requirementDef: RequirementDefinitionItem = {
                                     ...singleReqDef,
                                     icon: reqType.icon,
                                     requirementTypeId: reqType.id,
                                     requirementTypeTitle: reqType.title,
                                 };
-                                setRequirementDefinition(requirementDef);
+                                setRequirementDefinition(cloneRequirementDefinition(requirementDef)); //must clone here!
```
bratli, line 202: just a suggestion so you dont have to clone wtice is to use a seperate variable to be inserted into both.
const cloned = cloneRequirementDefinition(requirementDef);
setRequirementDefinition(cloned);
setNewRequirementDefinition(cloned);

bratli (approved):

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
arildbjerkenes: It is the original field. The one that is not modified.

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
arildbjerkenes: The original field (the one saved in the database) might be voided, but the user might have changed it to 'unvoid' without saving. Then, the delete button should not show, because that would give an error when trying to save. If we do the opposite,

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementDefinition.tsx
arildbjerkenes: NOPE! Dangerous. The requirementDefinition and newRequirementDefinition need to be separate objects, I couldn't use requirementDef either.

PR #238, opened 2020-08-31
77576 - Filter values not refreshed

kfaln:
Reset filters on project change. Handle refresh of scope list in correct component (ScopeOverview / ScopeFilter) based on current filter usage and visibility.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/ScopeFilter/ScopeFilter.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/ScopeFilter/ScopeFilter.tsx
@@ -279,10 +279,8 @@ const ScopeFilter = ({
     };
 
     useEffect(() => {
-        // On project change - handle reset of filters and refresh scope list (only when filters are active)
-        const activeFilters = Object.values(localTagListFilter).filter(v => v && JSON.stringify(v) != JSON.stringify([]));
-
-        if (activeFilters.length > 0 && projectNameRef.current !== project.name) {
+        // On project change - reset filters (triggers scope list update when filters were active)
+        if (projectNameRef.current !== project.name) {
             resetFilter();
         }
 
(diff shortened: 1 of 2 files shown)


src/modules/Preservation/views/ScopeOverview/ScopeFilter/ScopeFilter.tsx
```diff
@@ -277,6 +278,16 @@ const ScopeFilter = ({
...
     };
 
+    useEffect(() => {
+        // On project change - handle reset of filters and refresh scope list (only when filters are active)
+        const activeFilters = Object.values(localTagListFilter).filter(v => v && JSON.stringify(v) != JSON.stringify([]));
+
+        if (activeFilters.length > 0 && currentProjectName !== project.name) {
+            resetFilter();
+            setCurrentProjectName(project.name);
+        }
```
Cnordbo, line 288: Is there any downside to just shorttrack this and always just reset? (skip all the checking to see if the filter is already empty?)

Cnordbo: Same thing goes for project selection. Triggering the `setCurrentProjectName` will trigger a re-render, even after all the useEffects are run. Causing a re-render that doesnt actually have a repaint. IF we need this, we could/should(?) make use of the `useRef` instead, as this doesnt cause any re-renders.

Cnordbo (approved):

src/modules/Preservation/views/ScopeOverview/ScopeFilter/ScopeFilter.tsx
kfaln: Changed from state to ref. Removed check for active filters.
Both ScopeOverview and ScopeFilter components react on changes to the filters and we need to make sure that the query is triggered from the correct component. Was trying to avoid multiple queries being fired, but it works fine now without the active filter check.

PR #241, opened 2020-08-31
Always reset isLoading state

Cnordbo:
Endpoint returns a 404 when no entities exist. This caused the catch statement to trigger, and reset of isLoading state was skipped. 

DevOps#76455


src/modules/PlantConfig/views/Library/TagFunction/tabs/PreservationTab.tsx
```diff
@@ -69,6 +68,7 @@ const PreservationTab = (props: PreservationTabProps): JSX.Element => {
                 }
             }
         }
+        setIsLoading(false);
```
janhagevold, line 71: Could be added in a `finally `clause?

Cnordbo: It could, but does it provide any additional value, or would the result be the same? What do you think?

hagjan: It does not add any value. Just more explicit perhaps. Say if someone at a later stage wants to return earlier.

PR #243, opened 2020-08-31
Feature/75240 ui remove a tag from scope

arildbjerkenes:
Ideally, backend should have a bulk endpoint to remove tags. We don't have this so we just delete tag by tag.

### DIFF
--- a/src/modules/Preservation/views/EditTagProperties/EditTagProperties.tsx
+++ b/src/modules/Preservation/views/EditTagProperties/EditTagProperties.tsx
@@ -355,9 +355,11 @@ const EditTagProperties = (): JSX.Element => {
                 :
                 <Container>
                     <div>
-                        <ErrorContainer>
-                            {validationErrorMessage && (<Typography variant="caption">{validationErrorMessage}</Typography>)}
-                        </ErrorContainer>
+                        {validationErrorMessage &&
+                            <ErrorContainer>
+                                (<Typography variant="caption">{validationErrorMessage}</Typography>)
+                            </ErrorContainer>
+                        }
                         <InputContainer>
                             <SelectInput
                                 maxHeight={'300px'}
--- a/src/modules/Preservation/views/EditTagProperties/EditTagProperties.tsx
+++ b/src/modules/Preservation/views/EditTagProperties/EditTagProperties.tsx
@@ -1,8 +1,8 @@
-import { ButtonContainer, Container, Header, InputContainer, SpinnerContainer } from './EditTagProperties.style';
+import { ButtonContainer, Container, Header, InputContainer, SpinnerContainer, ErrorContainer } from './EditTagProperties.style';
 import { TagDetails, Step, Journey, RequirementType } from './types';
 import React, { useEffect, useRef, useState } from 'react';
 import SelectInput, { SelectItem } from '../../../../components/Select';
-import { Button } from '@equinor/eds-core-react';
+import { Button, Typography } from '@equinor/eds-core-react';
 import { usePreservationContext } from '../../context/PreservationContext';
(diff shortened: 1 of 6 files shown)


arildbjerkenes:
NOTE: I made a mistake, by not starting from master branch... I think it will work fine.

src/modules/Preservation/views/EditTagProperties/EditTagProperties.tsx
```diff
@@ -344,6 +355,9 @@ const EditTagProperties = (): JSX.Element => {
                 :
                 <Container>
                     <div>
+                        <ErrorContainer>
+                            {validationErrorMessage && (<Typography variant="caption">{validationErrorMessage}</Typography>)}
+                        </ErrorContainer>
```
Cnordbo, line 360: Add `ErrorContainer` to the conditional render.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -382,6 +382,50 @@ const ScopeOverview: React.FC = (): JSX.Element => {
...
+                rowVersion: t.rowVersion
+            })));
+            refreshScopeList();
+            setSelectedTags([]);
+            showSnackbarNotification('Selected tag(s) have been removed.');
+        } catch (error) {
+            console.error('Remove failed: ', error.message, error.data);
+            showSnackbarNotification(error.message);
+        }
+        return Promise.resolve();
```
Cnordbo, line 401: Not needed.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -382,6 +382,50 @@ const ScopeOverview: React.FC = (): JSX.Element => {
             completeFunc);
     };
 
+    let removableTags: PreservedTag[] = [];
+    let nonRemovableTags: PreservedTag[] = [];
```
Cnordbo, line 386: Make these local to `showRemoveDialog` and send them in to `remove` as parameters from the remove dialog.

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -382,6 +382,50 @@ const ScopeOverview: React.FC = (): JSX.Element => {
...
+        selectedTags.map((tag) => {
+            const newTag: PreservedTag = { ...tag };
+            if (tag.isVoided && !tag.isInUse) {
+                removableTags.push(newTag);
+            } else {
+                nonRemovableTags.push(newTag);
+            }
+        });
+        const removeButton = removableTags.length > 0 ? 'Remove' : null;
+        const removeFunc = removableTags.length > 0 ? remove : null;
```
Cnordbo, line 417: Solution when changing the "global" removableTags/nonremovabletags to local variables. 

```js
        const removeFunc = removableTags.length > 0 ? () => remove(removableTags) : null;
```

Cnordbo (changes requested):

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
arildbjerkenes: I know this is not a good solution, but all the other dialogs are implemented like this. The whole impl could need a refactoring. But we don't have time for this now. I think for this to work we would need to refactor showModelDialog, right? The callback functions do not have parameters.

Cnordbo: No, no large refactoring required, solution is proposed here: 

https://github.com/equinor/procosys-js-frontend/pull/243#discussion_r480030609

Cnordbo (approved):

PR #244, opened 2020-08-31
Feature/77826 active default filter

Cnordbo:
Saves and automatically loads the last triggered filter when navigating to Preservation. 

DevOps#77826

Additional: 

Refactors filter counting to be more consistent and automatically calculated.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
@@ -72,7 +72,7 @@ function setCachedFilter(projectId: number, filter: TagListFilter): void {
 function deleteCachedFilter(projectId: number): void {
     ScopeOverviewCache.delete(projectId + '-filter');
 }
-const defaultTagListFilter: TagListFilter = {
+const emptyTagListFilter: TagListFilter = {
     tagNoStartsWith: null,
     commPkgNoStartsWith: null,
     mcPkgNoStartsWith: null,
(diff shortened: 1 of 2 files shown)


src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -86,7 +115,16 @@ const ScopeOverview: React.FC = (): JSX.Element => {
     const [flyoutTagId, setFlyoutTagId] = useState<number>(0);
     const [scopeIsDirty, setScopeIsDirty] = useState<boolean>(false);
     const [pageSize, setPageSize] = useState<number>(50);
-    const [tagListFilter, setTagListFilter] = useState<TagListFilter>(defaultTagListFilter);
+    const [tagListFilter, setTagListFilter] = useState<TagListFilter>(() => {
+        const previousFilter = getCachedFilter(project.id);
+        if (previousFilter) {
+            return {
+                ...defaultTagListFilter,
```
kfaln, line 122: I realize now that the naming of `defaultTagListFilter` is conflicting with the filters the users can save as their default filter. This was made before default filters were a thing. The meaning here is "empty" or "initial" filter (= no filter). I suggest renaming to prevent confusion, perhaps "emptyTagListFilter".

Use of the actual default (user) filter remains to be implemented...

kfaln (changes requested):

src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
Cnordbo: Good idea, will look into this.

Cnordbo: renamed to `emptyTagListFilter`

kfaln (approved):

PR #260, opened 2020-09-07
Warning when project is closed

arildbjerkenes:
- A warning message will be shown, if a closed project is opened. 
- The only way to reach a closed project is to add project in url: http://.../Preservation?project=<project>
- Proper error handling when project does not exist, is implemented. 
- We need to fetch project from main-api, to find if it is closed. .


src/modules/Preservation/index.tsx
```diff
@@ -7,12 +7,15 @@ import { PreservationContextProvider } from './context/PreservationContext';
...
 import withAccessControl from '../../core/security/withAccessControl';
+import ClosedProjectWarning from './ClosedProjectWarning';
 
 const Preservation = (): JSX.Element => {
 
     const { url } = useRouteMatch();
+
     return (
         <PreservationContextProvider>
+            <ClosedProjectWarning />
```
kfaln, line 18: No check to see whether the project is **actually** closed? Seems like this is always rendered regardless of project status...

kfaln (changes requested):

src/modules/Preservation/index.tsx
arildbjerkenes: I was not finished with this. That's why I didn't put anyone as reviewer.

src/http/ProCoSysClient.ts
```diff
@@ -113,6 +120,24 @@ class ProCoSysClient extends ApiClient {
         return PascalCaseConverter.objectToCamelCase(result.data) as ProjectResponse[];
     }
 
+    /**
+     * Get project 
+     *
+     * @param setRequestCanceller Returns a function that can be called to cancel the request
+     */
+    async getProjectAsync(projectId: number, setRequestCanceller?: RequestCanceler): Promise<SingleProjectResponse> {
```
bratli, line 128: why cant we just use the Projects/{projectName} endpoint in preservation API?

arildbjerkenes: The preservation database only contains projects where tags are added to scope.

arildbjerkenes:
I have simplified the solution, to avoid complications.

src/modules/Preservation/ClosedProjectWarning.tsx
```diff
@@ -0,0 +1,45 @@
...
+const CloseProjectWarning = (): JSX.Element | null => {
+    const { procosysApiClient } = useProcosysContext();
+    const { project } = usePreservationContext();
+
+    const warningIcon = <EdsIcon name='warning_outlined' color={tokens.colors.interactive.warning__resting.rgba} size={24} />;
+
+    const [isProjectClosed, setIsProjectClosed] = useState<boolean>(false);
+
+    useEffect(() => {
+        let requestCancellor: Canceler | null = null;
```
Cnordbo, line 19: missing return statement in useEffect to cancel the request

src/modules/Preservation/ClosedProjectWarning.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { tokens } from '@equinor/eds-tokens';
+import { useProcosysContext } from '../../core/ProcosysContext';
+import { Canceler } from 'axios';
+import { showSnackbarNotification } from '@procosys/core/services/NotificationService';
+
+const CloseProjectWarning = (): JSX.Element | null => {
+    const { procosysApiClient } = useProcosysContext();
+    const { project } = usePreservationContext();
+
+    const warningIcon = <EdsIcon name='warning_outlined' color={tokens.colors.interactive.warning__resting.rgba} size={24} />;
```
Cnordbo, line 14: inline in return statement

src/modules/Preservation/ClosedProjectWarning.tsx
```diff
@@ -0,0 +1,45 @@
...
+                console.error('Get project closed statue failed: ', error.message, error.data);
+                showSnackbarNotification(error.message, 5000, true);
+            }
+        })();
+    }, [project.id]);
+
+    if (isProjectClosed) {
+        return (
+            <ClosedProjectContainer>
+                {warningIcon} <IconSpacer /> Selected project {project.name} is closed.
```
Cnordbo, line 38: Avoid using spacing components, as this will wrap differently on smaller screens. Rather wrap the text in a component that applies margins.

src/modules/Preservation/style.ts
```diff
@@ -0,0 +1,31 @@
...
+export const Container = styled.div`
+    padding: var(--margin-module--top) var(--margin-module--right);
+    display: flex;
+    height: 100%;
+    flex-direction: column;
+    h1 {
+        font-size: 36px;
+    }
+    overflow-y: scroll;
+`;
```
Cnordbo, line 13: i think your missing a merge from master. This is removed.

Cnordbo (changes requested):
A lot cleaner solution, good job. 

Just a few issues, and we are good to go :)

src/modules/Preservation/ClosedProjectWarning.tsx
arildbjerkenes: fixed

src/modules/Preservation/ClosedProjectWarning.tsx
arildbjerkenes: Hm, ok.

Cnordbo (approved):

PR #261, opened 2020-09-08
Feature/78078 scope overview improve responsivenes tmp

Cnordbo:
Initial responsivenes improvement for Scopeoverview. 

* Scrollable filter
* Scrollable table body
* Sticky table header
* Fixed overflows

Additional: 
* Switched H1 tags to use EDS Typography component 
* Import statements sorted in touched files.


bratli:
image

If I resize my window, the table is cropped.

bratli:
Scrollable filter, kan vi få scrollbar helt inntil siden?
image

src/modules/Preservation/index.tsx
```diff
@@ -1,8 +1,8 @@
 import { Route, BrowserRouter as Router, Switch, useRouteMatch } from 'react-router-dom';
 
 import AddScope from './views/AddScope/AddScope';
-import EditTagProperties from './views/EditTagProperties/EditTagProperties';
 import { Container } from './style';
```
bratli, line 4: remove

src/modules/Preservation/style.ts
```diff
@@ -1,13 +1,7 @@
 import styled from 'styled-components';
 
 export const Container = styled.div`
```
bratli, line 3: remove. Not in use

bratli (changes requested):
See previous comments with screenshots

arildbjerkenes:
Should the scroll-bar next to the table be thicker? At least, the scrollbars should have the same width.

arildbjerkenes (changes requested):
I approve all changes, except that I think we should verify the width on the scroll bars.

Cnordbo:
> I approve all changes, except that I think we should verify the width on the scroll bars.

No changes has been done on the width of scrollbars.

PR #270, opened 2020-09-10
Unable to edit Default Interval in Requirements

arildbjerkenes:
Following is fixed: 
- Edit tag does not work when journey is voided
- Styling issue with missing margin
- Issue with changing interval week 

I also included a two liner change on the Add scope manual search. Two characters are required to perform a search


kfaln (approved):

kfaln:
Check the failing test before completing.

PR #272, opened 2020-09-14
Use default filter on startup

arildbjerkenes:


src/modules/Preservation/views/ScopeOverview/ScopeFilter/SavedFilters.tsx
```diff
@@ -8,31 +8,24 @@ import { TagListFilter } from '../types';
 import EdsIcon from '@procosys/components/EdsIcon';
 import CloseIcon from '@material-ui/icons/Close';
 import { usePreservationContext } from '@procosys/modules/Preservation/context/PreservationContext';
+import { SavedFilter } from '../ScopeOverview';
```
Cnordbo, line 11: use a separate types file instead of exporting from another component.

src/modules/Preservation/views/ScopeOverview/ScopeFilter/ScopeFilter.tsx
```diff
@@ -396,7 +401,10 @@ const ScopeFilter = ({
                 }}
                 onClose={(): void => setShowSavedFilters(false)}
             >
-                <SavedFilters tagListFilter={tagListFilter}
+                <SavedFilters
+                    savedFilters={savedFilters}
```
Cnordbo, line 405: to avoid having a prop and component be named the same thing, this should probably just be named `filters` or `userFilters`

Cnordbo (changes requested):
Add tests

src/modules/Preservation/views/ScopeOverview/ScopeFilter/ScopeFilter.tsx
arildbjerkenes: Agree

PR #274, opened 2020-09-14
Add error boundary component

Cnordbo:
This adds a component that will prevent your other components or modules from crashing the entire application if an error occures. 

Usage: 
```js
<ErrorBoundary message="Optional custom error message" >
    <SuccessfullComponent />
</ErrorBoundary>
```

Important: 
Its always the mounting component that has to add the ErrorBoundary to the application. Not the component that the error is emerging from. 

So this does not work as a wrapper for every component, but rather as a wrapper that is added at the usage point. 

In the example above, that means that <SuccessFullComponent> can not have "its own protection" by just wrapping its return statement with <ErrorBoundary>

Example - will not work: 
```js
const myFaultFunction = () => throw 'error'

const SuccessfullComponent = () => {
    return (<ErrorBoundary><p>{myFaultyFunction()}</p></ErrorBoundary>
}
```

Example - will work: 
```js
const myFaultFunction = () => throw 'error'

const SuccessfullComponent = () => {
    return (<p>{myFaultyFunction()}</p>)
}

const ParentComponent = () => {
    return (<ErrorBoundary><SuccessfullComponent /></ErrorBoundary>)
}
```

### DIFF
--- a/src/components/ErrorBoundary/__tests__/ErrorBoundary.test.jsx
+++ b/src/components/ErrorBoundary/__tests__/ErrorBoundary.test.jsx
@@ -16,7 +16,7 @@ describe('<ErrorBoundary />', () => {
     console.error = jest.fn();
     it('Renders error with default message', () => {
         const { getByText } = render(<ErrorBoundary><FaultyComponent /></ErrorBoundary>);
-        expect(getByText('Sorry, an unexpected error occured')).toBeInTheDocument();
+        expect(getByText('An unexpected error occured')).toBeInTheDocument();
     });
 
     it('Renders error with custom error message', () => {
@@ -24,7 +24,7 @@ describe('<ErrorBoundary />', () => {
         expect(getByText('My custom error')).toBeInTheDocument();
     });
 
-    it('Renders error with custom error message', () => {
+    it('Renders component without error', () => {
         const { getByText } = render(<ErrorBoundary><SuccessfullComponent /></ErrorBoundary>);
         expect(getByText('Hello')).toBeInTheDocument();
     });
--- a/src/components/ErrorBoundary/__tests__/ErrorBoundary.test.jsx
+++ b/src/components/ErrorBoundary/__tests__/ErrorBoundary.test.jsx
@@ -1,0 +1,31 @@
+import ErrorBoundary from '..';
+import React from 'react';
+import { render } from '@testing-library/react';
+
+const FaultyComponent = () => {
+    throw 'This is an expected error from ErrorBoundary';
+};
+
+const SuccessfullComponent = () => {
+    return <p>Hello</p>;
+};
+
+
+describe('<ErrorBoundary />', () => {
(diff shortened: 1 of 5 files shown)


src/components/ErrorBoundary/__tests__/ErrorBoundary.test.jsx
```diff
@@ -0,0 +1,31 @@
...
+        const { getByText } = render(<ErrorBoundary><FaultyComponent /></ErrorBoundary>);
+        expect(getByText('Sorry, an unexpected error occured')).toBeInTheDocument();
+    });
+
+    it('Renders error with custom error message', () => {
+        const { getByText } = render(<ErrorBoundary message="My custom error"><FaultyComponent /></ErrorBoundary>);
+        expect(getByText('My custom error')).toBeInTheDocument();
+    });
+
+    it('Renders error with custom error message', () => {
```
kfaln, line 27: Duplicate (and faulty) description with the one above.

src/components/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+        super(props);
+        this.state = { hasError: false };
+    }
+
+    static getDerivedStateFromError(): ErrorState {
+        return { hasError: true };
+    }
+
+    componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
+        console.error('CRITICAL ERROR OCCURED');
```
kfaln, line 26: A bit dramatic? It might not be that critical :) "Unexpected" or "Unhandled" maybe?

src/components/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+        console.error('Error: ', error);
+        console.error('ErrorInfo: ', errorInfo);
+    }
+
+    render(): React.ReactChild {
+        if (this.state.hasError) {
+            return (
+                <ErrorBoundaryContainer>
+                    <Typography variant="h1">{!this.props.message ? 'Sorry, an unexpected error occured' : this.props.message}</Typography>
```
kfaln, line 35: Not sure how often we will provide the custom error message. So perhaps we should refine the default message a bit? Perhaps this is the place we put info like "please contact support if the problem persists bla bla"...

kfaln (changes requested):

src/components/ErrorBoundary/index.tsx
Cnordbo: Its think this only handles the most critical cases, as its an exception that will break the component and prevent it from rendering (which causes this to catch it). 

I think this is only caught if the component itself throws an exception, and not if a general promise is not caught.

src/components/ErrorBoundary/index.tsx
Cnordbo: Good idea.

kfaln (approved):

PR #278, opened 2020-09-16
Context and api client for ipo

paal123:
First end-point in place to retrive projects.

Note: Currently not using error handling, as we are refactoring the current PreservationApiError class


src/modules/PlantConfig/context/PlantConfigContext.tsx
```diff
@@ -18,17 +20,21 @@ export const PlantConfigContextProvider: React.FC = ({ children }): JSX.Element
     const { plant } = useCurrentPlant();
     const libraryApiClient = useMemo(() => new LibraryApiClient(auth), [auth]);
     const preservationApiClient = useMemo(() => new PreservationApiClient(auth), [auth]);
+    const invitationForPunchOutApiClient = useMemo(() => new InvitationForPunchOutApiClient(auth), [auth]);
```
bratli, line 23: Are we sure we need this here?

paal123: Correct! Removing it.

bratli (approved):

PR #281, opened 2020-09-18
Refresh scope list when default filter does not exist

arildbjerkenes:


arildbjerkenes:
Christer, if you are happy, just merge.

Cnordbo (approved):

PR #289, opened 2020-09-24
77729: summary view

bratli:


bratli:
@paal123 @arildbjerkenes  okay to just merge?

src/modules/InvitationForPunchOut/views/CreateIPO/Summary/Summary.tsx
```diff
@@ -0,0 +1,194 @@
...
+                </Table>
+            </TableSection>
+            <TableSection>
+                <Typography variant="h5">Selected scope</Typography>
+                <Table>
+                    <Head>
+                        { getHeaders() }
+                    </Head>
+                    <Body>
+                        { mcScope.length > 0 && mcScope.map(mcPkg => getMcScope(mcPkg)) }
```
paal123, line 151: should both be either getXXPkgScope or getXXScope?

paal123 (approved):
With suggestions

src/modules/InvitationForPunchOut/views/CreateIPO/Summary/Summary.tsx
bratli: Will fix this in next case since I have branched out from here :)

PR #292, opened 2020-09-29
77648:   participant view in wizard

bratli:


src/modules/InvitationForPunchOut/components/RoleSelector/index.tsx
```diff
@@ -0,0 +1,348 @@
...
+    }, containerRef);
+
+    const toggleDropdown = (): void => {
+        setIsOpen(!isOpen);
+    };
+
+    const clearOtherRoles = (roleIndex: number): void => {
+        setAllRoles(ar => {
+            const copyR = [...ar];
+            copyR.forEach((r, i) => {
```
paal123, line 84: Shold  we rather use more readable names? ar, copyR and r makes it a little  hard understand exactly what we are juggling here

src/modules/InvitationForPunchOut/views/CreateIPO/CreateIPO.tsx
```diff
@@ -178,7 +191,7 @@ const CreateIPO = (): JSX.Element => {
                 previous={goToPreviousStep}
                 participants={participants}
                 setParticipants={setParticipants}
-                isValid={true}
+                isValid={steps[2].isCompleted}
```
paal123, line 194: Could we create an Enum for steps to be able to use more readable names? Like Steps.Participants and a then at the same time be able to do if (step > 2) or if (step>Steps.Participants)

src/modules/InvitationForPunchOut/views/CreateIPO/Participants/Participants.tsx
```diff
@@ -1,36 +1,360 @@
...
+            return (): void => requestCanceler && requestCanceler();
+        } catch (error) {
+            showSnackbarNotification(error.message);
+        }
+    }, []);
+
+    useEffect(() => {
+        if(personFilter != '') {
+            try {
+                setFilteredPersons(testPart.filter(r => r.text.toLocaleLowerCase().startsWith(personFilter.toLocaleLowerCase())));
```
paal123, line 118: testPart?

src/modules/InvitationForPunchOut/views/CreateIPO/Participants/Participants.tsx
```diff
@@ -1,36 +1,360 @@
...
+        return JSON.parse(JSON.stringify(availableRoles));
+    };
+
+    const setOrganization = (value: string, index: number): void => {
+        setParticipants(p => {
+            const participantsCopy = [...p];
+            participantsCopy[index].organization = value;
+            return participantsCopy;
+        });
+        if(value == Organizations[6].value) {
```
paal123, line 137: Organizations as enum?

paal123 (approved):
With suggestions

src/modules/InvitationForPunchOut/views/CreateIPO/Participants/Participants.tsx
bratli: just using this until endpoint are ready - it is just dummy data. I have created a PBI to get persons from endpoint when this is ready

PR #294, opened 2020-09-30
Fix issue with collapse of treeview when tree is made dirty

arildbjerkenes:


src/components/TreeView/index.tsx
```diff
@@ -247,6 +267,26 @@ const TreeView = ({
...
+                for (const nodeId of pathToExpandTree) {
+                    const node = treeData.find(node => node.id === nodeId);
+                    if (node && !node.isExpanded) {
+                        await expandNode(node);
+                    }
+                }
+                setPathToExpandTree([]);
+            }
+            if (selectedNodeId) {
+                const selected = treeData.filter((node) => node.id === selectedNodeId);
```
kfaln, line 282: I guess there should only ever be one selected node, so perhaps use `treeData.find(...)` (instead of `filter`) and `if (selected)` below

kfaln (changes requested):

src/components/TreeView/index.tsx
arildbjerkenes: fixed

PR #303, opened 2020-10-05
In PO mode, preselect PO/call off no in add dummy tag

arildbjerkenes:


bratli:
PBI says that Supplier should be preselected as type. This is not implemented.

bratli (changes requested):
See previous comment

bratli (approved):

PR #304, opened 2020-10-06
Show spinner when loading saved filter

arildbjerkenes:


src/modules/Preservation/views/ScopeOverview/ScopeOverview.tsx
```diff
@@ -150,13 +152,15 @@ const ScopeOverview: React.FC = (): JSX.Element => {
     const [moduleAreaHeight, setModuleAreaHeight] = useState<number>(700);
 
     const updateSavedTagListFilters = async (): Promise<void> => {
+        setIsLoading(true);
```
kfaln, line 155: Is this the right place to set the loading state? I guess the loading has side-effects triggering scope list to be updated, so shouldn't the loading state rather be set where the scope update takes place? `refreshScopeList` ?

arildbjerkenes: Update of scope list will be triggered AFTER loading of the tag list filter. The problem is when it takes time for the tag list filter to load, there are no spinner indicating that things are loading.

kfaln (approved):

PR #305, opened 2020-10-06
Make requirement icons to link to tag flyout

arildbjerkenes:


kfaln (approved):
Ok, but please test that wrapping the icons in a new div don't have a negative impact on the rendering (margins, alignment in the row)

PR #306, opened 2020-10-07
Add force update on click for preservation requirements

janhagevold:


Cnordbo:
This issue is not just related to the preservation requirement, so the change should affect all library items. 

Remember to also check other options for forcing an update on the component. As toggling a variable true/false will/can cause a double render which is not needed. 

(useReducer, increment state by 1 e.e)

It could also be that we should see if there is a way to just change the active tree element to the newly created one, or just render back to the default view after creating the new entity, instead of showing the newly created entity within the same view.

Cnordbo (changes requested):
This issue is not just related to the preservation requirement, so the change should affect all library items.

Remember to also check other options for forcing an update on the component. As toggling a variable true/false will/can cause a double render which is not needed.

(useReducer, increment state by 1 e.e)

It could also be that we should see if there is a way to just change the active tree element to the newly created one, or just render back to the default view after creating the new entity, instead of showing the newly created entity within the same view.

src/modules/PlantConfig/views/Library/LibraryTreeview/LibraryTreeview.tsx
```diff
@@ -20,6 +22,9 @@ const LibraryTreeview = (props: LibraryTreeviewProps): JSX.Element => {
     } = usePlantConfigContext();
 
     const handleTreeviewClick = (libraryType: LibraryType, libraryItem: string): void => {
+        if (!libraryItem) {
+            props.forceUpdate();
```
arildbjerkenes, line 26: I think we should make it possible to use force update even if there is a library item.

arildbjerkenes (approved):

src/modules/PlantConfig/views/Library/LibraryTreeview/LibraryTreeview.tsx
janhagevold: At the moment, this will only re-render the component with no side effects. An idea would be that a forced re-render resets the component state, i.e. after editing to discard changes made without saving.

src/modules/PlantConfig/views/Library/Library.tsx
```diff
@@ -22,17 +22,19 @@ const Library = (): JSX.Element => {
...
 
     useEffect(() => {
-        setSelectedLibraryType(params.libraryType);
+        params.libraryType && setSelectedLibraryType(params.libraryType);
     }, []);
 
     return (
         <Container>
             <LibraryTreeview
+                forceUpdate={forceUpdate}
```
Cnordbo, line 37: is this needed? 
if state changes in this component, doesnt that just re-render all childs without having this referenced passed down?

PR #307, opened 2020-10-08
77708: attachment view in wizard

bratli:
-Can upload file by clicking button
-Can drag and drop files

### DIFF
--- a/src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
+++ b/src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
@@ -99,7 +99,7 @@ const Attachments = ({
                 actions={[
                     {
                         icon: (): JSX.Element => <EdsIcon name='delete_to_trash' />,
-                        tooltip: 'Delete User',
+                        tooltip: 'Remove attachment',
                         onClick: (_, rowData): void => removeAttachment(rowData.tableData.id)
                     }
                 ]}
--- a/src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
+++ b/src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
@@ -32,8 +32,9 @@ const Attachments = ({
     };
 
     const removeAttachment = (index: number): void => {
-        const newAttachmentsList = [...attachments.splice(index, 1)]; //todo: this is not necessary, but without anything here delete is not working
-        setAttachments([...attachments]);
+        const attachmentsCopy = [...attachments];
+        attachmentsCopy.splice(index, 1);
+        setAttachments(attachmentsCopy);
     };
 
     const getAttachmentName = (attachment: File): JSX.Element => {
--- a/src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
+++ b/src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
@@ -9,32 +9,28 @@ interface AttachmentsProps {
     next: () => void;
     previous: () => void;
     attachments: File[];
-    setAttachments: React.Dispatch<React.SetStateAction<File[]>>;
+    removeAttachment: (index: number) => void;
+    addAttachments: (attachments: File[]) => void;
 }
 
 const Attachments = ({
     next,
     previous,
     attachments,
-    setAttachments
+    removeAttachment,
+    addAttachments
 }: AttachmentsProps): JSX.Element => {
     const inputFileRef = useRef<HTMLInputElement>(null);
 
     const handleSubmitFile = (e: any): void => {
         e.preventDefault();
-        setAttachments([...attachments, e.target.files[0]]);
+        addAttachments([e.target.files[0]]);
     };
 
     const handleAddFile = (): void => {
         if (inputFileRef.current) {
             inputFileRef.current.click();
         }
-    };
-
-    const removeAttachment = (index: number): void => {
-        const attachmentsCopy = [...attachments];
-        attachmentsCopy.splice(index, 1);
-        setAttachments(attachmentsCopy);
     };
 
     const getAttachmentName = (attachment: File): JSX.Element => {
@@ -49,7 +45,7 @@ const Attachments = ({
     
     const handleDrop = (event: React.DragEvent<HTMLDivElement>): void => {
         event.preventDefault();
-        setAttachments(attachments.concat([...event.dataTransfer.files]));
+        addAttachments([...event.dataTransfer.files]);
     };
 
     return (<Container>
--- a/src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
+++ b/src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
@@ -1,20 +1,108 @@
-import React from 'react';
-import { Button } from '@equinor/eds-core-react';
-import { Container, FormContainer, ButtonContainer } from './Attachments.style';
+import React, { useRef } from 'react';
+import { Button, Typography } from '@equinor/eds-core-react';
+import { Container, FormContainer, ButtonContainer, AddAttachmentContainer, DragAndDropContainer } from './Attachments.style';
+import EdsIcon from '@procosys/components/EdsIcon';
+import { tokens } from '@equinor/eds-tokens';
+import Table from '@procosys/components/Table';
 
 interface AttachmentsProps {
     next: () => void;
     previous: () => void;
+    attachments: File[];
+    setAttachments: React.Dispatch<React.SetStateAction<File[]>>;
 }
 
 const Attachments = ({
     next,
-    previous
+    previous,
+    attachments,
+    setAttachments
 }: AttachmentsProps): JSX.Element => {
+    const inputFileRef = useRef<HTMLInputElement>(null);
+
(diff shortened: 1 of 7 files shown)


src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
```diff
@@ -1,20 +1,108 @@
...
+    };
+
+    const handleAddFile = (): void => {
+        if (inputFileRef.current) {
+            inputFileRef.current.click();
+        }
+    };
+
+    const removeAttachment = (index: number): void => {
+        const newAttachmentsList = [...attachments.splice(index, 1)]; //todo: this is not necessary, but without anything here delete is not working
```
bratli, line 35: @Cnordbo

src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
```diff
@@ -1,20 +1,109 @@
...
+    };
+    
+    const handleDragOver = (event: React.DragEvent<HTMLDivElement>): void => {
+        event.preventDefault();
+    };
+    
+    const handleDrop = (event: React.DragEvent<HTMLDivElement>): void => {
+        event.preventDefault();
+        const files = [...event.dataTransfer.files];
+        setAttachments([...attachments].concat(files));
```
Cnordbo, line 53: concat returns a new array, so the spread operator here is probably not needed. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
```diff
@@ -1,20 +1,109 @@
...
+        );
+    };
+    
+    const handleDragOver = (event: React.DragEvent<HTMLDivElement>): void => {
+        event.preventDefault();
+    };
+    
+    const handleDrop = (event: React.DragEvent<HTMLDivElement>): void => {
+        event.preventDefault();
+        const files = [...event.dataTransfer.files];
```
Cnordbo, line 52: You can use `event.dataTransfer.files` directly within concat.

src/modules/InvitationForPunchOut/views/CreateIPO/Summary/Summary.tsx
```diff
@@ -101,6 +103,14 @@ const Summary = ({
             </Row>);
     };
 
+    const getAttachments = (attachment: File, index: number): JSX.Element => {
```
Cnordbo, line 106: method naming, `getAttachments` - it just sounds like your getting the actual attachments, not just a single component. 

As well as pluralization of the name, when it only handles a single entity. 

This can also be created per render by just running
```js
const attachmentList = attachments.map((attachment, index) => (
<Row key={index}>
                <Cell>{attachment.name}</Cell>
            </Row>
));
```

And then render out the table based on the length of this directly. 

```js
{ attachmentList.length > 0 && <TableSection......> {attachmentList} </> }
```

src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
```diff
@@ -1,20 +1,109 @@
...
+import { Container, FormContainer, ButtonContainer, AddAttachmentContainer, DragAndDropContainer } from './Attachments.style';
+import EdsIcon from '@procosys/components/EdsIcon';
+import { tokens } from '@equinor/eds-tokens';
+import Table from '@procosys/components/Table';
 
 interface AttachmentsProps {
     next: () => void;
     previous: () => void;
+    attachments: File[];
+    setAttachments: React.Dispatch<React.SetStateAction<File[]>>;
```
Cnordbo, line 12: Instead of manipulating the state directly. What about creating methods that handle this at the parent level. 

`onAttachmentAdded`
`onAttachmentRemoved`

or just `onChange`

this way only the parent has control of what is actually is passed back down to the child element. and state is only mutated in a single place. Reducing the side effects of this component.

Cnordbo (changes requested):

src/modules/InvitationForPunchOut/views/CreateIPO/Attachments/Attachments.tsx
bratli: Would we send the Files to the parent component?

Also, do you think this is how it should be done in all components? I think we would end up with a lot of logic in the parent

bratli: Check it out if the last commit is what you meant :) I think this is nice!

Cnordbo: Yes, the list of Files would be sent to the parent, if we used the `onChange`. 
Otherwise, in the case of `onAttachmentAdded/removed` you would just send the element that is requested to be removed, or the element that is requested to be added. 

In any case the the parent would be responsible for updating the attachments state value based on that information.

In general, we should think of components as "self contained" meaning, that they only "possibly" alter states of parents or siblings based on events that happen in that component, and that the parent is responsible for reacting upon those events. 
Then the parent is responsible for submitting new data to the child which should be displayed. 
And only where it makes sense (like i believe it does here, since attachments doesnt really do any logic back and forth with the server, its just a tool for the parent component)

If this component would have had all the logic related to attachments within it (communicating with the server), I would not suggest the same approach as i am suggesting here. Altough this approach could be wrapped within another component that handles the server logic as well. 

ex. (with som fluff for illustration)

```js
<IHandleServerUploading>
   const {attachments, setAttachments} = useState<Attachment>([]);

   const upload = () => {
       //... Upload logic
   }

   const onAttachmentAdded = (newAttachment: Attachment) => setAttachments((existingAttachments) => 
   [...existingAttachments, newAttachment];

   const onAttachmentRemoved= (index: number) => setAttachments((existingAttachments) => {
     const newAttachmentList = [...existingAttachments];
     newAttachmentList.splice(index,1);
     return newAttachmentList;
   }

   <IHandleFileSelection_and_dragNDrop 
   allowedFileTypes={['jpg']} 
   dragAndDrop={false} 
   fileIcons={true} 
   attachments={attachments} 
   onAttachmentAdded={onAttachmentAdded}
   onAttachmentRemoved={onAttachmentRemoved}
    />

<button type="submit" onClick={upload} />
</IHandleServerUploading>
```

This has the following benefits: 
* We reduce the state mutation happening in child components of parent components (reducing side-effects)
* We can easily re-use the component later
* We are reducing the responsibility of the attachments component, or add more support to it later based on filetypes e.e

Hope this makes sense.

Cnordbo: with the onChange way of doing it, this would basically work the same, but then we pass back a copy of the attachments list, with the new file, or with the old file removed (a complete list) - and then the parent could either just set that directly as its own attachments value, or choose to validate before updating the list. (for instance in case of deleting an already uploaded file)

src/modules/InvitationForPunchOut/views/CreateIPO/Summary/Summary.tsx
```diff
@@ -206,9 +214,23 @@ const Summary = ({
                     </Body>
                 </Table>
             </TableSection>
-            <TableSection>
-                <Typography variant="h5">Attachments</Typography>
-            </TableSection>
+            { attachments.length > 0 &&
```
Cnordbo, line 217: You can use attachmentList here as well :)

src/modules/InvitationForPunchOut/views/CreateIPO/CreateIPO.tsx
```diff
@@ -147,14 +148,23 @@ const CreateIPO = (): JSX.Element => {
...
-        console.log(incompleteParticipantRows);
         if (incompleteParticipantRows.length > 0) {
             changeCompletedStatus(false, StepsEnum.Participants);
         } else {
             changeCompletedStatus(true, StepsEnum.Participants);
         }
     }, [participants]);
 
+    const removeAttachment = (index: number) => {
+        const attachmentsCopy = [...attachments];
```
Cnordbo, line 159: when manipulating state where we depend on the current value of the state, its best to use the callback function that can be sent with the setter. 

```js
setAttachments((currentState) => ....)
```

Cnordbo: This gives us a guarantee that the `attachment` state is actual current state when its being executed.

Cnordbo: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

src/modules/InvitationForPunchOut/views/CreateIPO/CreateIPO.tsx
```diff
@@ -147,14 +148,23 @@ const CreateIPO = (): JSX.Element => {
...
+        attachmentsCopy.splice(index, 1);
+        setAttachments(attachmentsCopy);
+    };
+
+    const addAttachments = (files: File[]) => {
+        setAttachments(attachments.concat(files));
```
Cnordbo, line 165: Same as previous comment regarding use of current state value.

Cnordbo (approved):
Have a look at the latest comments, PR can be merged after that. Good work!

PR #314, opened 2020-10-20
78632: get persons from api

bratli:
Get eprsons from API. Different endpoint depending on if it is the first contractor, construction company, or another participant. API call if the user has not typed for 300ms

### DIFF
--- a/src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
+++ b/src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
@@ -34,20 +34,21 @@ export interface McPkgResponse {
     disciplineCode: string;
 }
 
+export interface PersonResponse {
+    azureOid: string;
+    userName: string;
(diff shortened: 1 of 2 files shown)


paal123 (approved):

src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
```diff
@@ -34,20 +34,21 @@ export interface McPkgResponse {
     disciplineCode: string;
 }
 
+export interface PersonResponse {
```
Cnordbo, line 37: is export needed?

Cnordbo (approved):

src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
bratli: No, will remove in create ipo PR

PR #315, opened 2020-10-20
Feature/78950: create ipo

bratli:


src/modules/InvitationForPunchOut/views/CreateIPO/Participants/Participants.tsx
```diff
@@ -23,12 +23,12 @@ enum OrganizationsEnum {
 
 const Organizations: SelectItem[] = [
     { text: OrganizationsEnum.Commissioning, value: OrganizationsEnum.Commissioning },
-    { text: OrganizationsEnum.ConstructionCompany, value: OrganizationsEnum.ConstructionCompany },
+    { text: OrganizationsEnum.ConstructionCompany, value: 'ConstructionCompany' },
```
paal123, line 26: For consistency we should probably rather use OrganizationEnum[OrganizationEnum.ConstructionCompany] instead of a mix of hardcoded and enum values

src/modules/InvitationForPunchOut/views/CreateIPO/CreateIPO.tsx
```diff
@@ -24,14 +30,14 @@ const emptyGeneralInfo: GeneralInfoDetails = {
...
-        organization: 'Contractor',
+        organization: { text: OrganizationsEnum.Contractor, value: OrganizationsEnum.Contractor },
         type: 'Functional role',
         externalEmail: null,
         person: null,
         role: null
     },
     {
-        organization: 'Construction company',
+        organization: { text: OrganizationsEnum.ConstructionCompany, value: 'ConstructionCompany' },
```
paal123, line 40: OrganizationEnum[OrganizationsEnum.ConstructionCompany]

paal123 (approved):
With suggestions

src/modules/InvitationForPunchOut/views/CreateIPO/Participants/Participants.tsx
bratli: The reason why I am using some hardcoded values is to satisfy the enum backend. Let me know if you have a good solution on how to handle this best...

bratli: Will add it to another card to find a solution to this so that I can complete this PR

src/modules/InvitationForPunchOut/views/CreateIPO/CreateIPO.tsx
bratli: Will add it to another card to find a solution to this so that I can complete this PR

PR #327, opened 2020-10-30
Feature 78031:   UI Attachment tab

janhagevold:
# Description

- Add getAttachments and deleteAttachment api
- Add Attachment tab
- Cleanup and minor fixes Participant table

Fixes # DevOps#78031

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have linked my DevOps task with this pull-request


Cnordbo (changes requested):

src/modules/InvitationForPunchOut/views/ViewIPO/Attachments/index.style.ts
```diff
@@ -0,0 +1,67 @@
...
+    width: 100%;
+    > * {
+        margin-bottom: calc(var(--grid-unit) * 3);
+    }
+`;
+
+export const ButtonContainer = styled.div`
+    display: flex;
+    justify-content: flex-end;
+    width: 100%;
```
Cnordbo, line 33: In my local testing, it looks like i can remove all the `width: 95-100%` properties from this file, without affecting the rendered results. 
Do you see an effect? 

Point is, make sure we only use properties that has an impact on the rendered results.

src/modules/InvitationForPunchOut/views/ViewIPO/Attachments/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+    };
+
+    useEffect(() => {
+        const getAttachmentsOnMount = async (): Promise<void> => {
+            setLoading(true);
+            const response = await getAttachments();
+            setAttachments(response);
+            setLoading(false);
+        };
+        getAttachmentsOnMount();
```
Cnordbo, line 49: This never cancels its own request. Setup a request canceler that gets returned from the useEffect to be triggered on unload.

Cnordbo (approved):

PR #365, opened 2020-12-01
Bug 79944: UI - Entering  0 hours in AM/PM 

janhagevold:
# Description

It appears that when entering 0 hours when locale is AM/PM, the date input returns an empty string. 
Added check: the resulting array "HH:mm" is non-empty for hours and minutes.

Completes: [AB#79944](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/79944)

# Checklist:

### DIFF
--- a/src/modules/InvitationForPunchOut/views/CreateIPO/__tests__/CreateIPO.test.tsx
+++ b/src/modules/InvitationForPunchOut/views/CreateIPO/__tests__/CreateIPO.test.tsx
@@ -3,6 +3,7 @@ import { fireEvent, render, waitFor, within } from '@testing-library/react';
 import CreateIPO from '../CreateIPO';
 import React from 'react';
 import { act } from 'react-dom/test-utils';
+import { configure } from '@testing-library/dom';
 
 jest.mock('react-router-dom', () => ({
     useParams: (): {projectId: any, commPkgNo: any} => ({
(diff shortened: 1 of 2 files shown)


src/modules/InvitationForPunchOut/views/CreateIPO/__tests__/CreateIPO.test.tsx
```diff
@@ -59,6 +60,35 @@ describe('<CreateIPO />', () => {
         const { getByText } = render(<CreateIPO />);
         await waitFor(() => expect(getByText('Previous').closest('button')).toHaveProperty('disabled', true));
     });
+
+    it('Should update end time when entering start time > end time', async () => {
+        configure({ testIdAttribute: 'id' }); // makes id attibute data-testid for subsequent tests
```
Cnordbo, line 65: this is the only place where `configure` is setup. 
Should this be put into the describe method instead? So that the test arent relying on a specific order of execution?

Cnordbo (approved):
Approved with comment. Merge when resolved.

src/modules/InvitationForPunchOut/views/CreateIPO/__tests__/CreateIPO.test.tsx
janhagevold: Using data-testid in preceeding test. It can also be put outside the test also to affect all.

PR #376, opened 2020-12-07
Feature 78826:  UI show attachment type icon

janhagevold:
# Description

- Add Type field for attachment tables with icon
- Add tests for utility functions

Completes: [AB#78826](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/78826)

# Checklist:

### DIFF
--- a/src/modules/InvitationForPunchOut/views/utils.tsx
+++ b/src/modules/InvitationForPunchOut/views/utils.tsx
@@ -8,6 +8,7 @@ export const getFileTypeIconName = (fileName: string): string => {
         case 'JPEG':
         case 'PNG':
         case 'SVG':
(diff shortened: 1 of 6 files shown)


src/modules/InvitationForPunchOut/views/utils.tsx
```diff
@@ -0,0 +1,41 @@
+
+
+export const getFileTypeIconName = (fileName: string): string => {
```
Cnordbo, line 3: Filepath, this is not a view. So it should probably be moved outside the views path.

Cnordbo (approved):
see feedback

PR #387, opened 2020-12-18
Bug 80194: UI only show complete/approve for first Contractor/ConstructionCompany

janhagevold:
# Description

Change behaviour in participants table. Contractor and ConstructionCompany gets Complete/Approve based on sortKey order.

Completes: [AB#80195](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/80195)

# Checklist:


bratli (approved):
I think we should do the same for the first operation, comissioning, and TI.
So if there are one of each, then each can have the sign property. If there are three TI, then only the first has sign.
If we have Supplier, commissioning, and Operation, then commissioning and operation will have sign.

Let me know if it is unclear. Decide if you want to do it in this PR or create a new PBI (I am working on backend canSign-property, so it can wait until that is done)

PR #395, opened 2020-12-29
Feature 77747: UI - List comments

janhagevold:
# Description

- Add Comments pane
- Add endpoint get post comment
- Add tests
- Add add comments

Added the add comments code to this PR from [[AB#77746](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/77746)].

Completes: [AB#77747](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/77747)

# Checklist:


src/modules/InvitationForPunchOut/views/ViewIPO/types.d.ts
```diff
@@ -66,3 +66,14 @@ export type Invitation = {
...
+
+export type IpoComment = {
+    id: number;
+    comment: string;
+    createdAtUtc: string;
+    createdBy: {
+        firstName: string;
+        lastName: string;
+    },
+    rowVersion: string;
```
bratli, line 78: I dont think we need rowVersion after all, because we will not have to option to remove a comment. So you can remove rowVersion from the type. You can change it in the next PR for addComment if nothing else to change here :)

bratli (changes requested):
I think that the comments flyout should have its own scroll, if that makes sense? So if we have 50 comments, I think we should be able to scroll the comments without scrolling the general info

bratli:
We can talk to Jon as well if you disagree/are unsure about comment scrolling

bratli:
Sorry, but I dont think this looks very good. Can we use filter on preservation as inspiration

image

bratli:
Another example:
image

bratli:
I think that the comments should take the whole length of the component. NwoNow the comments are cut off half way

image

bratli:
Text is overflowing the comments. Maybe have the header part of comment fixed when scrolling comments?

image

bratli:
image
The cursor does not change when hovering this button. I think it should, as the cursor changes when hovering all other forms of buttons. Maybe even have the background change on hover, like many other places

bratli:
image
grey separator does not go all the way to the bottom

bratli:
Comments should be open by default

bratli:
@arildbjerkenes talked with Jon. We need to change the scrolling on "View IPO" on bot the "main component" and "comment component" for better UI. So the "main part" will have one scrollbar, and "comments" will have one scrollbar. Progressbar will be fixed.

bratli (changes requested):
Comments should be open by default

bratli (changes requested):
Comments should be open by default

src/modules/InvitationForPunchOut/views/ViewIPO/index.tsx
```diff
@@ -40,6 +47,43 @@ const ViewIPO = (): JSX.Element => {
     const { apiClient } = useInvitationForPunchOutContext();
     const [invitation, setInvitation] = useState<Invitation>();
     const [loading, setLoading] = useState<boolean>(false);
+    const [showComments, setShowComments] = useState<boolean>(false);
```
bratli, line 50: basically set this to true I think

bratli (approved):

PR #396, opened 2020-12-29
Feature 78032: UI - Log tab

janhagevold:
# Description

- Add log tab
- Add test
- Add endpoint

Completes: [AB#78032](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/78032)

# Checklist:


bratli:
Change name from Log to History, as this is how it is for Preservation

src/modules/InvitationForPunchOut/views/ViewIPO/types.d.ts
```diff
@@ -66,3 +66,19 @@ export type Invitation = {
     mcPkgScope: McPkgScope[];
     commPkgScope: CommPkgScope[];
 }
+
+export type HistoryItem = {
+    id: number;
+    description: string;
+    createdAtUtc: string;
+    createdBy: {
```
bratli, line 74: same here, remove unused person-attributes

bratli (changes requested):

src/modules/InvitationForPunchOut/views/ViewIPO/History/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+                    <Spinner large />
+                </SpinnerContainer>
+            ) : (
+                <FormContainer>
+                    <HistoryTable>
+                        <Head>
+                            <Row>
+                                <Cell as="th" scope="col" style={{verticalAlign: 'middle'}} width="15%">Date</Cell>
+                                <Cell as="th" scope="col" style={{verticalAlign: 'middle'}} width="10%">Changed by</Cell>
+                                <Cell as="th" scope="col" style={{verticalAlign: 'middle'}} width="10%">Event type</Cell>
```
bratli, line 62: I dont think we should show eventType - this is not in the design.

On all endpoints, where may be some fields on that are not necessarily used in UI, but are added because the developer thinks that if other parts of equinor uses the endpoint, the information may be nice for them

bratli (approved):

PR #408, opened 2021-01-08
Feature 80318: UI - Text before IPO created

janhagevold:
# Description

- Add checkbox and confirmation text to general info 
- Add check for checkbox checked to proceed

Completes: [AB#80318](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/80318)

# Checklist:


src/modules/InvitationForPunchOut/views/CreateAndEditIPO/GeneralInfo/GeneralInfo.tsx
```diff
@@ -215,6 +219,19 @@ const GeneralInfo = ({
                     }}
                 />
             </LocationContainer>
+            <ConfirmationTextContainer>
+                <ConfirmCheckbox checked={confirmationChecked} onChange={(): void => setConfirmationChecked(confirmed => !confirmed)} />
```
bratli, line 223: I think having this checkbox in edit mode is unneccessary. Either remove it from edit mode, or have it as default checked (and unable to be unchecked)

bratli: You can ask Jon if you want, but for me either is ok. If Jon asked you to do it like this then okay

bratli (changes requested):

src/modules/InvitationForPunchOut/views/CreateAndEditIPO/GeneralInfo/GeneralInfo.tsx
janhagevold: Added disabled checkbox on edit IPO. Seems eds checkbox does not register disabled when controlled..

bratli (approved):

PR #412, opened 2021-01-14
Cancel IPO button

arildbjerkenes:
# Description

Cancel IPO button.

Completes: [AB#79515](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/79515)

# Checklist:

### DIFF
--- a/src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
+++ b/src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
@@ -718,13 +718,14 @@ class InvitationForPunchOutApiClient extends ApiClient {
     *
     * @param setRequestCanceller Returns a function that can be called to cancel the request
     */
-    async cancelPunchOut(id: number, setRequestCanceller?: RequestCanceler): Promise<void> {
+    async cancelPunchOut(id: number, rowVersion: string, setRequestCanceller?: RequestCanceler): Promise<void> {
         const endpoint = `/Invitations/${id}/Cancel`;
         const settings: AxiosRequestConfig = {};
         this.setupRequestCanceler(settings, setRequestCanceller);
         try {
-            const result = await this.client.post(
+            const result = await this.client.put(
                 endpoint,
+                { rowVersion: rowVersion },
                 settings
             );
             return result.data;
--- a/src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
+++ b/src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
@@ -714,6 +714,26 @@ class InvitationForPunchOutApiClient extends ApiClient {
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
```diff
@@ -713,6 +713,26 @@ class InvitationForPunchOutApiClient extends ApiClient {
         }
     }
 
+    /**
+    * Cancel PunchOut
+    *
+    * @param setRequestCanceller Returns a function that can be called to cancel the request
+    */
+    async cancelPunchOut(id: number, setRequestCanceller?: RequestCanceler): Promise<void> {
```
bratli, line 721: There will be a change to the cancel endpoint, but need to wait for main API to be deployed to QP. Do you want to just hold this PR until that is ready?
The change is that we send inn a dto, RowVersion, which is the row version of the invitaiton

src/modules/InvitationForPunchOut/views/ViewIPO/ViewIPOHeader.tsx
```diff
@@ -31,6 +34,18 @@ const ViewIPOHeader = (props: ProgressBarProps): JSX.Element => {
         setDisplayFlyout(true);
     };
 
+    const confirmCancelIpo = (): void => {
+        showModalDialog(
+            'Cancel IPO',
+            <div>This will cancel the IPO and send a cancellation to invited participants.</div>,
+            '18vw',
+            'Cancel',
```
bratli, line 42: Is if confusing to have Cancel in the cancel dialogue? Dont know what would be better... Maybe confirm with Jon

src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
bratli: I think it is best to wait until the web api is in qp, because if not, we will suddenly have a cancel button frontend that does not work - but up to you

src/modules/InvitationForPunchOut/views/ViewIPO/ViewIPOHeader.tsx
bratli: Maybe "Close" or "Exit"?

arildbjerkenes: Good point. Will ask Jon.

src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
bratli: @arildbjerkenes the changes backend are merged. So send row version as well now

bratli (approved):
Nice

PR #417, opened 2021-01-19
Update signedBy field based on changes from backend

arildbjerkenes:
# Description
Update signedBy field based on changes from backend
Completes: [AB#80413](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/80413)

# Checklist:

### DIFF
--- a/src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
+++ b/src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
@@ -48,7 +48,15 @@ type ParticipantInvitationResponse = {
     externalEmail: ExternalEmailInvitationResponse;
     person: PersonInvitationResponse;
     functionalRole: FunctionalRoleInvitationResponse;
-    signedBy?: string;
+    signedBy?: {
+        id: number;
+        firstName: string;
+        lastName: string;
+        userName: string;
+        azureOid: string;
+        email: string;
+        rowVersion: string;
+    },
     signedAtUtc?: Date;
     attended: boolean;
     note: string;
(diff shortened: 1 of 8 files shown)


src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
```diff
@@ -48,7 +48,15 @@ type ParticipantInvitationResponse = {
     externalEmail: ExternalEmailInvitationResponse;
     person: PersonInvitationResponse;
     functionalRole: FunctionalRoleInvitationResponse;
-    signedBy?: string;
+    signedBy?: {
```
bratli, line 51: I think you should only take the "attributes" that you need here, and you can get the others later if you need them

bratli (changes requested):
You need to do the same for CreatedBy on invitation that is shown in the outlook flyout

src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
```diff
@@ -41,14 +41,32 @@ type CommPkgScopeResponse = {
     status: string;
 }
 
+type CreatedByInvitationResponse = {
```
bratli, line 44: Why not just make a personResponse and use it on both createdBy and signedBy? They are the same dto backend fyi

src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
```diff
@@ -146,11 +114,13 @@ interface McPkgResponse {
 }
 
 interface PersonResponse {
```
bratli, line 116: This has been used somewhere else where id and row version is not a part of the response. From what we talked about earlier I think you agree that this requires two seperate interfaces? One for PersonResponse as it was before (used other places), and a new one (call it something else) for this. Or am I missing something?

bratli (changes requested):

PR #424, opened 2021-01-22
Switch to Azure App Configuration for application settings

Cnordbo:
# Description

## BREAKING CHANGE
This sets up the application to fetch its configuration from Azure App Configuration. 
All users needs to change their `settings.json` to reflect these changes. 
Local overrides can added where needed, by manipulating the `settings.json` file with the appropriate properties. 
Unreleased settings should be mapped, with the feature PR, in ProCoSys settings, and the Azure App Configuration service. 

## Dependencies: 
- Need to setup the production environment for app configuration
- Frontend pipeline needs to refactor how its setting configuration variables
- Azure key vault to hold track of endpoint and endpoint scope

Completes: [AB#79969](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/79969)

# Checklist:

Tests will be added, but as this is a big review, so I am publishing before these are implemented.


Cnordbo:
Build failure is expected as the pipeline uses the old configuration setup.

src/modules/UserGreeting/index.tsx
```diff
@@ -18,6 +18,7 @@ const UserGreeting = (): JSX.Element => {
     let imageRequestToken: null | Canceler = null;
     const graphClient = new GraphClient(auth);
 
+
```
bratli, line 21: remove line (will that remove history for this file?)

src/http/SettingsApiClient.ts
```diff
@@ -0,0 +1,66 @@
...
+
+interface FeatureConfig {
+    url: string
+    scope: Array<string>
+    version?: string
+}
+
+interface ConfigResponse {
+    configuration: {
+        procosysApi: FeatureConfig
```
bratli, line 13: rename to proCoSysMainApi? or proCoSysWebApi? Dont know if this may become confusing for someone at some point

settings.template.json
```diff
@@ -1,42 +1,4 @@
...
-            "version": ""
-        },
-        "libraryApi": {
-            "url": "",
-            "scope": "[]",
-            "version": ""
-        }
-    },
-    "instrumentationKey": ""
+    "configurationEndpoint": "",
```
bratli, line 2: do we still need this file?

src/auth/AuthService.tsx
```diff
@@ -60,11 +47,21 @@ export default class AuthService implements IAuthService {
     private authInstance: UserAgentApplication;
 
     constructor() {
+        const authConfig: Configuration = {
+            auth: {
+                clientId: ProCoSysSettings.clientId,
+                redirectUri: window.location.href,
+                authority: ProCoSysSettings.authority,
+            },
+            // system: {
```
bratli, line 56: should this commented out bit be romoved?

src/app/index.tsx
```diff
@@ -1,32 +1,27 @@
...
+import { UserContextProvider } from '@procosys/core/UserContext';
+import withFeatureFlag from '@procosys/core/features/withFeatureFlag';
 
-const Login = React.lazy(() => import('../modules/Login'));
 const GeneralRouter = React.lazy(() => import('./GeneralRouter'));
 
 const App = (): JSX.Element => {
-    const {auth} = useProcosysContext();
-
-    const userIsLoggedIn = auth.getCurrentUser() != null;
```
bratli, line 15: is this done somewhere else, or do we not need it?

src/app/index.tsx
```diff
@@ -1,32 +1,27 @@
...
+                        <GeneralRouter />
+                    </Suspense>
+                </ThemeProvider>
+            </MuiThemeProvider>
+        </UserContextProvider>
     );
 };
 
-export default App;
+export default withFeatureFlag(App, ['main']);
```
bratli, line 27: BTW, are the necessay feature flags / variables added in Azure so things are ready there?

settings.template.json
Cnordbo: Yes, we use it for setting up which API the new configuration is received from. As well as provide developers with an override option for each of the settings.

Cnordbo: The template file is used by the build server, and submitted as a boilerplate to the open source repository.

src/app/index.tsx
Cnordbo: Only for dev environment atm. Will be added to test\prod ASAP.

src/app/index.tsx
Cnordbo: This is currently refactored to first entrypoint `index.tsx`:
https://github.com/equinor/procosys-js-frontend/blob/tech/79827_Use_azure_app_configuration/src/index.tsx#L37-L46

src/auth/AuthService.tsx
Cnordbo: Short: Its just there to make our life easier when we need to debug Authentication, and as its a part of MSAL its just there to inform us that this is an available option that can help with debugging such scenarios. 

Just for information - This is in current master as well. So its not added\new. 
( https://github.com/equinor/procosys-js-frontend/blob/master/src/auth/AuthService.tsx#L12-L14 ). 
(Not that thats an excuse, just info) 

We can remove it if you want.

src/http/SettingsApiClient.ts
Cnordbo: Its a good idea, The naming here was just based on what the old property was.

src/modules/UserGreeting/index.tsx
Cnordbo: Yes, that should remove the history saying that this file was affected by this PR. (if we do a squash commit instead of merging)

src/http/SettingsApiClient.ts
Cnordbo: A good idea, the naming is based on what the name is in the old property. We should create a task to actually refactor this instead, as ProCoSysApi is used everywhere today (ProCoSysClient, and the same name is used inn all modules that rely on Procosys Main API) 

If i refactor this now, that will cause an additional 20 files to be added to the list of changes here. 
So i would suggest that we do this as a separate step in the february sprint. 

Does this sound like a good solution to you?

bratli: Sounds good

bratli (approved):

PR #442, opened 2021-02-03
Feature 80735: UI - Table of IPOs

janhagevold:
# Description

- Add material-table from Preservation design

Completes: [AB#80735](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/80735)

# Checklist:


src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
```diff
@@ -150,6 +151,30 @@ interface FunctionalRoleResponse {
     persons: ParticipantPersonResponse[];
 }
 
+interface IPO {
+    id: number;
+    title: string;
+    type: string;
+    status: string;
+    commPkgs?: Pick<CommPkgResponse, 'commPkgNo'>[];
```
bratli, line 159: just fyi, when the endpoint is ready I am just returing a list of strings for commPkgNos and mcPkgNos, as we dont need as much detail in this endpoint

src/modules/InvitationForPunchOut/http/InvitationForPunchOutApiClient.ts
```diff
@@ -253,6 +278,61 @@ class InvitationForPunchOutApiClient extends ApiClient {
...
+
+        this.setupRequestCanceler(settings, setRequestCanceller);
+
+        // try {
+        //     const result = await this.client.get<IPOsResponse>(endpoint, settings);
+        //     return result.data;
+        // } catch (error) {
+        //     throw new IpoApiError(error);
+        // }
+        return await new Promise<IPOsResponse>(resolve => setTimeout(() => resolve(
```
bratli, line 311: Same here, when endpoint is ready you will see some different naming of attributes than this, becuase I have tried to follow the same stle as for the rest of IPO. Eg. Completed is CompeltedAtUtc

src/modules/Header/index.tsx
```diff
@@ -179,6 +179,11 @@ const Header: React.FC = (): JSX.Element => {
                             >
                                 <DropdownItem>Hookup types</DropdownItem>
                             </a>
+                            { (ProCoSysSettings.featureIsEnabled('IPO')) &&
```
bratli, line 182: Dont merge this part in before everything is ready. This is so that users in QP/prod wont find it and think that is is ready. For now we can just type the url when testing

bratli, line 182: Also, is it possible to make it a little bit longer? The scrollbar is jjust scrolling some millimeters, so I would rather not have it 
image

src/modules/InvitationForPunchOut/views/SearchIPO/Table/index.tsx
```diff
@@ -0,0 +1,177 @@
+import { CommPkg, IPO, McPkg } from '../types';
```
bratli, line 1: I think it is important that we follow the same design on tables. This table is "cleaner" than the ones in preservation feks, or the ones in IPO when searching for scope (not as many horizaontal lines). If you think this design is better, talk to Jon to see if he agrees, and then create a PBI to make all tables in this design. It is important that we follow the same design, if not the next developer will add his preferences, and we will end up with a ton of different designs. 
To summarize, either make this table follow the design of other tables, or create a PBI to change other tables in new solution to follow this design

bratli (changes requested):

src/modules/Header/index.tsx
janhagevold: The height depends on the view height with default 80vh. We can definitely bump this up a bit if your screen is a standard that we would see often.

src/modules/InvitationForPunchOut/views/SearchIPO/Table/index.tsx
janhagevold: I had added a line of css to remove the very bottom border as it looked cleaner. Didn't notice that it affected all the rows. Reverting this for now so it corresponds with existing tables.

arildbjerkenes (approved):

PR #443, opened 2021-02-04
Feature 80725: UI - Disable Cancel button

janhagevold:
# Description

- Add history.back() functionality to Cancel button from create IPO
- Reimplement dirty state for edit IPO

Completes: [AB#80725](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/80725)

# Checklist:

### DIFF
--- a/src/modules/InvitationForPunchOut/views/CreateAndEditIPO/CreateIPO.tsx
+++ b/src/modules/InvitationForPunchOut/views/CreateAndEditIPO/CreateIPO.tsx
@@ -99,10 +99,19 @@ const CreateIPO = (): JSX.Element => {
     const [availableRoles, setAvailableRoles] = useState<RoleParticipant[]>([]);
     const { apiClient } = useInvitationForPunchOutContext();
     const { history } = useRouter();
-    const { unsetDirtyStateFor } = useDirtyContext();
     const [steps, setSteps] = useState<Step[]>(initialSteps);
     const [projectNameFromMain] = useState<string | null>(params.projectName ? decodeURIComponent(params.projectName) : null);
     const [commPkgNoFromMain] = useState<string | null>(params.commPkgNo ? decodeURIComponent(params.commPkgNo) : null);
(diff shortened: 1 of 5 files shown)


bratli:
Was it difficult to get the Cancel button to just go back to where you came from? I think we especially need this when we will soon have table view for IPOs, and the user can click "Create new IPO" from this page. Then the cancel-button should go back to that page. I think we should always go back to the page we were on before? Or should we always go back to IPO table view when that is ready?

bratli:
ALSO, didnt we have a warning when leaving an IPO that is not saved? It seems to be gone now

bratli (changes requested):
See comments.

Jon wanted to use history to go back

src/modules/InvitationForPunchOut/views/CreateAndEditIPO/CreateIPO.tsx
```diff
@@ -99,11 +99,20 @@ const CreateIPO = (): JSX.Element => {
...
     const { history } = useRouter();
-    const { unsetDirtyStateFor } = useDirtyContext();
     const [steps, setSteps] = useState<Step[]>(initialSteps);
     const [projectNameFromMain] = useState<string | null>(params.projectName ? decodeURIComponent(params.projectName) : null);
     const [commPkgNoFromMain] = useState<string | null>(params.commPkgNo ? decodeURIComponent(params.commPkgNo) : null);
 
+    const { setDirtyStateFor, unsetDirtyStateFor } = useDirtyContext();
+
+    useEffect(() => {
+        if (JSON.stringify(generalInfo) !== JSON.stringify(initialGeneralInfo)) {
```
bratli, line 109: Is this to check if the next / save button should be enabled?

Did we have anything before that gave the user a warning if they left an unsaved page?

bratli (approved):

PR #448, opened 2021-02-11
Bug 80407: Can not click attachments before create

janhagevold:
# Description

- Add URL.createObjectURL on file object to retrieve file before upload

Completes: [AB#80407](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/80407)

# Checklist:

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,3 +1,7 @@
 import '@testing-library/jest-dom/extend-expect';
 
-window.URL.createObjectURL = function() {};
+function noOp () { }
+
+if (typeof window.URL.createObjectURL === 'undefined') { 
+  Object.defineProperty(window.URL, 'createObjectURL', { value: noOp})
+}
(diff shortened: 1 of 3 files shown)


jest.setup.js
```diff
@@ -1 +1,3 @@
 import '@testing-library/jest-dom/extend-expect';
+
+window.URL.createObjectURL = function() {};
```
Cnordbo, line 3: There is an ongoing issue about this in jsdom, so to avoid overwriting this at any point where this is implemented, we should probably wrap it. 

```js
function noOp () { }
if (typeof window.URL.createObjectURL === 'undefined') { 
  Object.defineProperty(window.URL, 'createObjectURL', { value: noOp})
}
```

https://github.com/jsdom/jsdom/issues/1721

Cnordbo (changes requested):

Cnordbo (approved):

PR #451, opened 2021-02-11
Feature 80941: UI - Add application insights custom events

janhagevold:
# Description

- Add custom events for created, edited, completed, accepted, signed, canceled

Completes: [AB#80941](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/80941)

# Checklist:

### DIFF
--- a/src/core/PlantContext.tsx
+++ b/src/core/PlantContext.tsx
@@ -89,6 +89,7 @@ export const PlantContextProvider: React.FC = ({ children }): JSX.Element => {
 
     useEffect(() => {
         setCurrentPlant(plantInPath);
+        analytics.setCurrentPlant(currentPlant.title);
     }, [plantInPath]);
 
     if (!currentPlant) {
--- a/src/core/PlantContext.tsx
+++ b/src/core/PlantContext.tsx
@@ -89,7 +89,7 @@ export const PlantContextProvider: React.FC = ({ children }): JSX.Element => {
 
     useEffect(() => {
         setCurrentPlant(plantInPath);
-        analytics.setCurrentPlant(currentPlant.title);
+        analytics.setCurrentPlant(plantInPath);
     }, [plantInPath]);
 
     if (!currentPlant) {
--- a/src/core/PlantContext.tsx
+++ b/src/core/PlantContext.tsx
@@ -60,6 +60,7 @@ export const PlantContextProvider: React.FC = ({ children }): JSX.Element => {
         const plant = plantsFiltered[0] as PlantContextDetails;
         plant.pathId = normalizedPlantId;
         cache.setCache('plant', plant);
+        analytics.setCurrentPlant(plant.id);
         setCurrentPlantInContext(plant);
     };
 
@@ -72,7 +73,6 @@ export const PlantContextProvider: React.FC = ({ children }): JSX.Element => {
 
     useEffect(() => {
         procosysApiClient.setCurrentPlant(currentPlant.id);
-        analytics.setCurrentPlant(currentPlant.id);
         let requestCanceler: Canceler;
         (async (): Promise<void> => {
             try {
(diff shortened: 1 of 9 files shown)


src/modules/InvitationForPunchOut/views/CreateAndEditIPO/CreateIPO.tsx
```diff
@@ -262,6 +266,7 @@ const CreateIPO = (): JSX.Element => {
                     mcPkgScope,
                     commPkgScope
                 );
+                analystics.trackUserAction(IpoCustomEvents.CREATED, { plant: plant.title, project: generalInfo.projectName, type: generalInfo.poType.value });
```
Cnordbo, line 269: Just to close a  tech story (not yet created) at the same time here. 

We should extend the Analytics context with a function for setting the plant, and then change that plant from the plantcontext everytime current plant changes. 

plantContext already uses analytics, so it should be a simple win, with a potentially huge gain for us. And you would avoid sending in the plant for every event you want to track. 

As Project and other parameters arent default or required by every module, this still needs to be sent in as you do here. 

This will help us setup so that all events potentially has a plant (if specified)

Simply a request, but doing this later requires a small refactor of your code here.

src/modules/InvitationForPunchOut/views/CreateAndEditIPO/EditIPO.tsx
```diff
@@ -266,6 +270,7 @@ const EditIPO = (): JSX.Element => {
                     commPkgScope,
                     invitation.rowVersion
                 );
+                analystics.trackUserAction(IpoCustomEvents.EDITED, { plant: plant, project: generalInfo.projectName, type: generalInfo.poType.value });
```
Cnordbo, line 273: this is sending in the entire plant object, where as other places your using `plant.title`

Cnordbo (changes requested):
Fix the `plant` vs `plant.title` comment. Everything else looks fine. 

Take into consideration the universal plant setup in analytics if you wish. or else we should create a tech story on this.

src/core/PlantContext.tsx
```diff
@@ -89,6 +89,7 @@ export const PlantContextProvider: React.FC = ({ children }): JSX.Element => {
 
     useEffect(() => {
         setCurrentPlant(plantInPath);
+        analytics.setCurrentPlant(currentPlant.title);
```
Cnordbo: Should we use the ID here? It might be easier to track across projects then. (PCS$JOHAN_SVERDRUP)

janhagevold: changed to plantInPath

src/core/PlantContext.tsx
```diff
@@ -89,6 +89,7 @@ export const PlantContextProvider: React.FC = ({ children }): JSX.Element => {
 
     useEffect(() => {
         setCurrentPlant(plantInPath);
+        analytics.setCurrentPlant(plantInPath);
```
Cnordbo, line 92: this should react to 
```js
useEffect(() => { 
  analytics.setCurrentPlant(currentPlant.id);
}, [currentPlant])
```

Cnordbo (approved):
Address that last issue, and we should be good to go. 

LGTM.

Cnordbo (approved):

src/core/PlantContext.tsx
janhagevold: With dependency on currentPlant, the analytics instance gets plant on first page load. When changing plant, _plant on the instance is set to empty string (constructor is executed again?).
_plant is updated properly as intended when moved analytics.setCurrentPlant() to PlantContext's own setCurrentPlant function as in latest commit.

Cnordbo (approved):

PR #460, opened 2021-02-25
Feature 88132: UI - Filter adjustments

janhagevold:
# Description

- Add adjustments to search table. Display invitation on single row with ellipsis for long text and tooltip.
- Add DateService formatting follow browser settings. Use DateService in all components and tests.
- Fix dirty state is only unset for one element when several should be cleared (Add 'unsetDirtyStateForMany' to context)
- Fix bug where attachment dirty state is set incorrectly in create and edit IPO
- Fix inconsequent typography in IPO tables
- Fix ugly console error caused by nested a tags in search table (<a><Typography link>)
- Fix attachment tab not actually returning objectURL for files.
- Add Helmet page titles for IPO and Preservation

Completes: AB#88132

# Checklist:

### DIFF
--- a/src/core/services/DateService.ts
+++ b/src/core/services/DateService.ts
@@ -1,47 +1,59 @@
 
+// THE COMMENTED FUNCTIONS HAVE BEEN USED BY PRESERVATION
+// KEEP FOR CONVENIENCE UNTIL EDS DATETIME STRATEGY IS IN PLACE
+
 /**
  * Return formatted date (dd.mm.yyyy). Empty string is returned if date is null.
  * @param date 
  */
-export const getFormattedDate = (date: Date | null): string => {
-    if (date === null) {
-        return '';
-    }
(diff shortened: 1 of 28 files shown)


src/core/services/DateService.ts
```diff
@@ -1,47 +1,59 @@
...
-    if (date === null) {
-        return '';
-    }
-    const newDate = new Date(date);
-    const d = newDate.getDate();
-    const m = newDate.getMonth() + 1;
-    const y = newDate.getFullYear();
-    return '' + (d <= 9 ? '0' + d : d) + '.' + (m <= 9 ? '0' + m : m) + '.' + y;
-};
+// export const getFormattedDate = (date: Date | null): string => {
```
bratli, line 9: did you want to keep these commented functions in case we need to revert something?

bratli (approved):
See comment

PR #488, opened 2021-03-18
Fix global dirty handling for library

arildbjerkenes:
# Description

Global handling if dirty state is fixed with this PR. Library is updated to use the global dirty handling. 

Completes: [AB#74997](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/74997)

# Checklist:

### DIFF
--- a/src/components/TreeView/index.tsx
+++ b/src/components/TreeView/index.tsx
@@ -4,6 +4,7 @@ import KeyboardArrowDownIcon from '@material-ui/icons/KeyboardArrowDown';
 import KeyboardArrowRightIcon from '@material-ui/icons/KeyboardArrowRight';
 import { TreeContainer, NodeContainer, ExpandCollapseIcon, NodeName, NodeLink } from './style';
 import Spinner from '../Spinner';
+import { unsavedChangesConfirmationMessage, useDirtyContext } from '@procosys/core/DirtyContext';
 
 /**
  * @param id Unique identifier across all nodes in the tree (number or string).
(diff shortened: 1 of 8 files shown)


src/components/TreeView/index.tsx
```diff
@@ -40,7 +41,7 @@ const TreeView = ({
     resetDirtyNode
 }: TreeViewProps): JSX.Element => {
 
-
+    const { isDirty } = useDirtyContext();
```
Cnordbo, line 44: TreeView is a component, and should not know about which context its loaded from. `isDirty` should be passed as a prop to the TreeView Component to render the prompt on select.

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementType.tsx
```diff
@@ -56,20 +57,39 @@ const PreservationRequirementType = (props: PreservationRequirementTypeProps): J
         setIconList(items);
     }, []);
 
+    const isDirty = (): boolean => {
+        return JSON.stringify(requirementType) != JSON.stringify(newRequirementType);
```
Cnordbo, line 61: This can be a `useMemo` that reacts to changes in `newJourney` state value. 
Using `useMemo` could help with performance as well, as it doesnt get recalculated everytime this is called. JSON.stringify is considered a slow function in javascript. 

This also changes some of the other implementations of isDirty in this module, so adapt accordingly.

Cnordbo (changes requested):

src/modules/PlantConfig/views/Library/PreservationRequirements/PreservationRequirementType.tsx
arildbjerkenes: I started to do this refactoring, but because isDirty is only called when there are changes to the form, and when buttons are clicked I didn't do it. In Mode isDirty is used in the rendering, so there is it more important. But anyway, I will do the refactoring. Makes it more consistent and clean.

PR #490, opened 2021-03-22
75541/UI   hover state på checkbox

VRiveland:
# Description

Changed the ProCoSys checkbox to look like the EDS one. Fixed Alignment and positioning issues caused by the checkbox having changed in size because of the hover state.

Completes: [AB#75541](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/75541)

# Checklist:

### DIFF
--- a/src/modules/InvitationForPunchOut/views/SearchIPO/Filter/SavedFilters/index.tsx
+++ b/src/modules/InvitationForPunchOut/views/SearchIPO/Filter/SavedFilters/index.tsx
@@ -127,7 +127,7 @@ const SavedFilters = (props: SavedFiltersProps): JSX.Element => {
                     />
                 </div>
 
-                < div style={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems:'center' }}>
+                <div style={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems:'center' }}>
                     <Checkbox
                         checked={newFilterIsDefault}
                         onChange={(checked: boolean): void => {
@@ -140,7 +140,7 @@ const SavedFilters = (props: SavedFiltersProps): JSX.Element => {
                         Save filter
                     </Button>
                 </div>
-            </Container >
+            </Container>
         );
     }
 
@@ -182,13 +182,12 @@ const SavedFilters = (props: SavedFiltersProps): JSX.Element => {
                                         {deleteIcon}
                                     </Button>
                                 </div>
-
                             </Row>
                         </React.Fragment>
                     );
                 })}
             </ListContainer>
-        </Container >
+        </Container>
     );
 };
 
--- a/src/modules/InvitationForPunchOut/views/SearchIPO/Filter/SavedFilters/index.tsx
+++ b/src/modules/InvitationForPunchOut/views/SearchIPO/Filter/SavedFilters/index.tsx
@@ -127,7 +127,7 @@ const SavedFilters = (props: SavedFiltersProps): JSX.Element => {
                     />
                 </div>
 
-                < div style={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between' }}>
+                < div style={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems:'center' }}>
                     <Checkbox
                         checked={newFilterIsDefault}
                         onChange={(checked: boolean): void => {
--- a/src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
+++ b/src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
@@ -29,7 +29,6 @@ const moduleName = 'PreservationJourneyForm';
 const WAIT_INTERVAL = 300;
 
 const checkboxHeightInGridUnits = 4;
-const checkboxWrapperMaxHeight = 'calc(var(--grid-unit) * 4)';
 
 enum AutoTransferMethod {
     NONE = 'None',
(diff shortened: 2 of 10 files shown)


src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
```diff
@@ -773,7 +773,7 @@ const PreservationJourney = (props: PreservationJourneyProps): JSX.Element => {
                                     {index != 100 &&
                                         <div style={{ fontSize: '12px', paddingBottom: 'var(--grid-unit)' }}>Automatic transfer on signing</div>
                                     }
-                                    <div style={{ display: 'flex', flexDirection: 'row' }}>
+                                    <div style={{ display: 'flex', flexDirection: 'row', maxHeight: 'calc(var(--grid-unit) * 4)' }}>
```
paal123, line 776: Since you are doing a change in this file. Please update all
 showSnackbarNotification(error.message, 5000);
to
 showSnackbarNotification(error.message);

since it is a default value.

paal123, line 776: Use a  const instead of hardcoded value

src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
```diff
@@ -785,6 +785,7 @@ const PreservationJourney = (props: PreservationJourneyProps): JSX.Element => {
                                                         setStepAutoTransValue(AutoTransferMethod.NONE, index);
                                                     }
                                                 }}
+                                                heightInGridUnits={4}
```
paal123, line 788: Use a  const instead of hardcoded value

src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
```diff
@@ -801,6 +802,7 @@ const PreservationJourney = (props: PreservationJourneyProps): JSX.Element => {
...
```
paal123, line 805: Use a  const instead of hardcoded value

src/modules/Preservation/views/ScopeOverview/ScopeFilter/CheckboxFilter.tsx
```diff
@@ -1,11 +1,12 @@
 import React, { useState } from 'react';
-import { Collapse, CollapseInfo, Section, ExpandedContainer } from './ScopeFilter.style';
+import { Collapse, CollapseInfo } from './ScopeFilter.style';
 import { Typography } from '@equinor/eds-core-react';
 import KeyboardArrowDownIcon from '@material-ui/icons/KeyboardArrowDown';
 import KeyboardArrowUpIcon from '@material-ui/icons/KeyboardArrowUp';
 import Checkbox from '../../../../../components/Checkbox';
 import { CheckboxFilterValue, TagListFilterParamType } from './ScopeFilter';
 import EdsIcon from '@procosys/components/EdsIcon';
+import { ExpandedContainer } from './CheckboxFilter.style';
```
paal123, line 9: Move it up to keep it next for the other imported style (line 2)

src/modules/InvitationForPunchOut/views/SearchIPO/Filter/SavedFilters/index.tsx
```diff
@@ -127,7 +127,7 @@ const SavedFilters = (props: SavedFiltersProps): JSX.Element => {
                     />
                 </div>
 
-                < div style={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between' }}>
+                < div style={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems:'center' }}>
```
paal123, line 130: < div <-- remove space

paal123 (changes requested):
Since you are affecting the CheckBox component used several places, I guess you have verified that all other usages of the checkbox component is not affected negatively?

VRiveland:
> Since you are affecting the CheckBox component used several places, I guess you have verified that all other usages of the checkbox component is not affected negatively?

Since I've only changed the look of the component, I haven't checked the functionality everywhere. I verified that the checkbox works a few places, and I can't imagine how it could not work everywhere else unless it already didn't. I could however check everywhere if you feel like it's needed.
I did however check that the checkbox looks correct everywhere.

paal123:
> > Since you are affecting the CheckBox component used several places, I guess you have verified that all other usages of the checkbox component is not affected negatively?
> 
> Since I've only changed the look of the component, I haven't checked the functionality everywhere. I verified that the checkbox works a few places, and I can't imagine how it could not work everywhere else unless it already didn't. I could however check everywhere if you feel like it's needed.
> I did however check that the checkbox looks correct everywhere.

Regression testing is an important part of the developmentprosess since your changes is not tested by anybody else before it hits productions, meaning you have to be confident that your changes does not have a negative effect on existing functionality.

In 100% of cases where production fails due to unforseen side effect, we never can imagine, understand or foresee how our changes can break related functionality, thats why we do regression testing to minimize the possibility.

Based on my experience I strongly recommend you do the regression test, but the change it self look ok so I approve the pull request, but it is your responsibility as the developer that you do the necessary testing (functional and regression) of your functionality before you complete the pull request.

paal123 (approved):
Approve but strongly advise you to do the regression test before completing.

src/modules/PlantConfig/views/Library/PreservationJourney/PreservationJourney.tsx
VRiveland: There are several places in this file where a hardcoded value is used instead of a const inside an inline style, and there are several places somewhere else in the code where maxHeight is hardcoded inside an inline style. Is there a reason why those are there and this cannnot?

I will absolutely change this of course, but it would be nice to know if there are any rules for this for future reference. E.g. padding can be hardcoded inline, but maxHeight cannot or something like that.

PR #491, opened 2021-03-22
bug/81916 Users without the ipo_sign privilege group can be added as …

VRiveland:
…a participant in signer organizations

# Description

Persons without the ipo_sign privilege group can no longer be added as a person participant in an organization type that can sign an IPO even if the organization isn't the first of it's kind. 

Completes: [AB#81916](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/81916)

# Checklist:

### DIFF
--- a/src/modules/InvitationForPunchOut/views/CreateAndEditIPO/Participants/Participants.tsx
+++ b/src/modules/InvitationForPunchOut/views/CreateAndEditIPO/Participants/Participants.tsx
@@ -237,11 +237,11 @@ const Participants = ({
     };
 
     const isSignerParticipant = (index: number): boolean => {
-        if ((participants[index].organization.value == Organizations[0].value ||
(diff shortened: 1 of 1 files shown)


src/modules/InvitationForPunchOut/views/CreateAndEditIPO/Participants/Participants.tsx
```diff
@@ -237,17 +237,11 @@ const Participants = ({
     };
 
     const isSignerParticipant = (index: number): boolean => {
-        if(index < 2){
-            return true;
-        }
-        if (index < 5 && (participants[index].organization.value == Organizations[0].value ||
+        if ((participants[index].organization.value == Organizations[0].value ||
```
bratli, line 240: I think there is an extra set of () that can be removed here

bratli (approved):

PR #511, opened 2021-04-29
Tech/80641 attachment component

VRiveland:
# Description

Attachment lists in IPO now also uses the AttachmentList component. Attachment list in ViewIPO should now look like the other tables, and not like the old design.
A drag and drop field has been added to attachment components in tag flyout in preservation to match design, and it is now possible to upload more than one file at a time like it is in IPO. File type icon has also been added to match design.

Completes: [AB#80641](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/80641)

# Checklist:


VRiveland:
There is an error when you first render an attachment component in tag flyout in preservation, this stems from before these changes. I'll be investigating more where it originates from, but it seems to have appeared around the time when the new table component was introduced.

janhagevold (approved):

PR #600, opened 2022-01-07
Bug/88829 ipo cannot use more than one signing button without refreshing

VRiveland:
# Description

AttNoteData and CleanData wasn't updated with the new rowVersion when one of the signing buttons or the update button was used, this caused an issue when backend started checking row versions properly.

The AttNoteData and CleanData are now updated whenever the participant list is updated. The participant list is updated when the invitation is updated, and the invitation was already updated whenever one of the signing endpoints or the update endpoint was called.

Completes: [AB#88829](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/88829)

# Checklist:


chormaz (approved):
Ser ok ut.

prestholm2 (approved):
👍

PR #605, opened 2022-01-18
Feature/89019 change ipo signed by to name

VRiveland:
# Description

Changed the "signed by" value in the participants table to the name of the user instead of the shortname.
Removed tests for rendering of signed by value, as they don't add anything useful. (The way we write tests should be redone in this repo to be more like in the apps.)

Changed to using the id given in the participant in the get invitations/{id} endpoint, as the id used (on person, role or external email) is deprecated.

Added type checking on error in catch in the EditIPO file. This is needed after a change in typescript. If you don't like the way this was handled, or you know of an even better way then leave a comment.

Completes: [AB#89019](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/89019)

# Checklist:


prestholm2 (approved):
👍

PR #704, opened 2022-12-16
76108/preservationUI-Voidede-tags-er-grå-i-transfer-dialog

AnneSkje:
# Description

Added textColour as optional colour in DialogTable and standard as black

Completes: [AB#76108](https://statoildeveloper.visualstudio.com/b6d97460-2e53-47ae-aad4-b41396d90828/_workitems/edit/76108)

# Checklist:

- [ ] I have performed a self-review of my own code.
- [ ] I have linked my DevOps task using the AB# tag.
- [ ] My code is easy to read, and comments are added where needed.

### DIFF
--- a/src/modules/Preservation/views/ScopeOverview/Dialogs/SharedCode/DialogTable.tsx
+++ b/src/modules/Preservation/views/ScopeOverview/Dialogs/SharedCode/DialogTable.tsx
@@ -3,24 +3,35 @@ import { PreservedTag } from '../../types';
 import { Container } from './DialogTable.style';
 import ProcosysTable from '@procosys/components/Table';
 
-interface TableProps {
+interface RequiredTableProps {
     tags: PreservedTag[];
     columns: any[];
     toolbarText?: string;
     toolbarColor?: string;
 }
 
+interface OptionalTableProps {
+    textColor: string;
+}
+
+interface TableProps extends RequiredTableProps, OptionalTableProps {}
+
+const defaultProps: OptionalTableProps = {
+    textColor: 'black',
+};
+
 const DialogTable = ({
     tags,
     columns,
     toolbarText,
     toolbarColor,
+    textColor,
 }: TableProps): JSX.Element => {
     const numTags = tags.length;
 
     return (
         <Container>
-            <div>
+            <div style={{ color: textColor }}>
                 <ProcosysTable
                     columns={columns}
                     data={tags}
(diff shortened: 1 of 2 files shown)


src/modules/Preservation/views/ScopeOverview/Dialogs/SharedCode/DialogTable.tsx
```diff
@@ -3,24 +3,35 @@ import { PreservedTag } from '../../types';
...
 
-interface TableProps {
+interface RequiredTableProps {
     tags: PreservedTag[];
     columns: any[];
     toolbarText?: string;
     toolbarColor?: string;
 }
 
+interface OptionalTableProps {
```
VRiveland, line 13: If you'r going to make a distinction between optional and required props then you should also move the other optional props into this interface.

The way we usually do this is to just use one interface with all props, and then assign a default value when creating the component

src/modules/Preservation/views/ScopeOverview/Dialogs/SharedCode/DialogTable.tsx
```diff
@@ -3,24 +3,35 @@ import { PreservedTag } from '../../types';
...
+const defaultProps: OptionalTableProps = {
+    textColor: 'black',
+};
+
 const DialogTable = ({
     tags,
     columns,
     toolbarText,
     toolbarColor,
+    textColor,
```
VRiveland, line 28: textColor = black, 
Here will do the same as your code above, as long as you add textColor?: string to table props

src/modules/Preservation/views/ScopeOverview/Dialogs/SharedCode/DialogTable.tsx
```diff
@@ -3,24 +3,35 @@ import { PreservedTag } from '../../types';
...
 }: TableProps): JSX.Element => {
     const numTags = tags.length;
 
     return (
         <Container>
-            <div>
+            <div style={{ color: textColor }}>
```
VRiveland, line 34: No inline style

VRiveland: This makes the text in the whole table the color you input. Even the header row. This is probably not wanted.
Additionally, as you can see in the description above the table for non-transferable tags, tags in this table can also be not started or already completed. Those should not be grey.

VRiveland (changes requested):

VRiveland (approved):

PR #815, opened 2024-03-20
72397  bug/72397 library framework should support url that will bring up the page

kslazykv:
AB#72397


chormaz:
Seems to work - but a couple of comments:

1.
image

Links in the treeview should be the same color as other links (i.e. not blue)

2.
When I right-click on an item in the list and select open in new tab, it displays the correct item. However, the treeview is not expanded to the selected item. 
image

chormaz (changes requested):
See comment in PR

larskay (approved):

PR #841, opened 2024-07-31
bug/109038 When "Punch out to" is empty user receives error when exporting to excel

kslazykv:
[AB#109038](https://statoildeveloper.visualstudio.com/Statoil.NextGenProCoSys/_boards/board/t/Statoil.NextGenProCoSys%20Team/Backlog%20items/?workitem=109038)


src/modules/InvitationForPunchOut/views/SearchIPO/Filter/InvitationsFilter.tsx
```diff
@@ -300,6 +300,7 @@ const InvitationsFilter = ({
     };
 
     const handleExportToExcel = async () => {
+        const { punchOutDateFromUtc, punchOutDateToUtc } = localFilter;
```
larskay, line 303: You declare these variables 2 times, can't you lift them out of the functions and only declare them once?

PR #917, opened 2025-02-14
fix: library header item

Gustav-Eikaas:
Fixed a bug where `Plant configuration` header item would redirect to `PlantConfig` locally instead of `libraryv2`
Added env var to allow for yellow header in procosystest


skagebouvet (approved):
🚀

PR #928, opened 2025-03-04
disable button if sufficient permissions

sondreravndal:
# Description

Added a tooltip and disabled button if user do not have permission

[Completes: AB#FILL_IN_YOUR_ISSUE_ID
](https://github.com/equinor/cc-toolbox/issues/1257#issuecomment-2696548630)

# Checklist:

- [ ] I have performed a self-review of my own code.
- [ ] I have linked my DevOps task using the AB# tag.
- [ ] My code is easy to read, and comments are added where needed.

### DIFF
--- a/src/modules/Preservation/views/AddScope/AddScope.tsx
+++ b/src/modules/Preservation/views/AddScope/AddScope.tsx
@@ -354,10 +354,30 @@ const AddScope = (): JSX.Element => {
                 error.message,
                 error.data
             );
+
+            let errorMessage = 'An unexpected error occurred.';
+            if (error.data) {
+                switch (error.data.status) {
+                    case 401:
+                        errorMessage = 'You are not authorized';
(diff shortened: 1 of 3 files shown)


src/modules/Preservation/views/AddScope/AddScope.tsx
```diff
@@ -354,10 +354,30 @@ const AddScope = (): JSX.Element => {
...
+                    case 403:
+                        errorMessage =
+                            'You do not have permission to access this resource.';
+                        break;
+                    case 404:
+                        errorMessage = 'The requested resource was not found.';
+                        break;
+                    case 500:
+                        errorMessage =
+                            'An unexpected error occurred. Please try again later.';
```
Gustav-Eikaas, line 373: Probably best to add a default case at bottom instead of 500

